PR #9, opened 2015-03-25
Factor somatic variant callers in the same GADT variant

smondet:
This should make it easier to add somatic variant callers.
(i.e. this kind of dev but somatic
https://github.com/hammerlab/biokepi/compare/add-haplotype-caller)

- a new Biokepi variant caller would just have to create a function in `Biokepi_pipeline.Construct`
- anyone can create that kind of step without having to modify the library (as long as they comply with the API in `Somatic_variant_caller.t`)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/9)

### DIFF
--- a/src/app/main.ml
+++ b/src/app/main.ml
@@ -36,7 +36,7 @@ let crazy_somatic_example ~normal_fastqs ~tumor_fastqs ~dataset =
           mutect bam_pair;
           somaticsniper bam_pair;
           somaticsniper ~prior_probability:0.001 ~theta:0.95 bam_pair;
-          varscan bam_pair;
+          varscan_somatic bam_pair;
         ])
   in
   vcfs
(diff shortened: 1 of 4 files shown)


arahuja:
Cool this seems like a good idea! I felt like I was copy+pasting a lot of code

arahuja:
General question - what happens if certain somatic callers don't support multiple processors? It seems like some of the caller just drop the processors at some point?

smondet:
Good question, I was thinking of parametrizing the number of processors in `Biokepi_run_environment.Machine.t` as “Upper bound preference” for a given machine/cluster.

Then a when a tools can do multi-threading it would query that number from the `Machine.t`.

arahuja:
Yea, I think an upper bound makes sense. Right now, do individual variant callers run in separate containers or a single one?

smondet:
Avery "target" runs separately (with Yarn as backend, yes, it goes to a
different container)
(when mutect is "map-reduced" of chromosomes, it makes 25 simultaneous
targets/containers)

PR #11, opened 2015-03-30
add sorting and indexing of bams as dependency to strelka

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/11)

### DIFF
--- a/src/lib/biokepi_target_library.ml
+++ b/src/lib/biokepi_target_library.ml
@@ -792,6 +792,8 @@ The usage is:
       Machine.get_reference_genome run_with `B37 |> Reference_genome.fasta in
     let strelka_tool = Machine.get_tool run_with "strelka" in
     let gatk_tool = Machine.get_tool run_with "gatk" in
+    let sorted_normal = Samtools.sort_bam ~run_with normal in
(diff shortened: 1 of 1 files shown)


src/lib/biokepi_target_library.ml
```diff
@@ -792,6 +792,8 @@ The usage is:
       Machine.get_reference_genome run_with `B37 |> Reference_genome.fasta in
     let strelka_tool = Machine.get_tool run_with "strelka" in
     let gatk_tool = Machine.get_tool run_with "gatk" in
+    let sorted_normal = Samtools.sort_bam ~run_with normal in
+    let sorted_tumor = Samtools.sort_bam ~run_with tumor in
     let make =
       Machine.run_program run_with ~name ~processors
         Program.(
```
smondet: You should use `sorted_normal` and `sorted_tumor` on lines 811, and 812, also right?

```
...
 normal#product#path
tumor#product#path
tumor#product#path
reference_fasta#product#path
reference_fasta#product#path
...
```

(Github does not let me comment on those "context" lines)

arahuja: Ah yes .. as the input of course

PR #15, opened 2015-03-31
Add suport for GRCh38 and hg19

arahuja:
- Added download of GRCh38 and hg19
- Made reference build passed to the variant callers configurable

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/15)


arahuja:
I updated this with a few things:
- using existing `fasta` files rather than concat-ing (to avoid GATK complaining)
- allowing the mapreduce functions to know the major contigs of each reference
- adding additional dependencies to strelka to have a fasta dictionary for combinevariants

This works for strelka and for mutect (except that the current mutect requires cosmic which we don't have for hg19 so mutect won't allow you to use hg19.  I think is fine for now - removing the cosmic input works and I will a hg19 cosmic file (if you know of a link let me know, otherwise I will use the `LiftoverVariants` tool as suggested by GATK

PR #16, opened 2015-03-31
Fix samtools sort flag

arahuja:
`samtools` was not happy with the quoting that happened

```
'samtools' 'sort -@ 1' '/hpc/users/ahujaa01/dream/synthetic.challenge.set2.normal.bam' '/hpc/users/ahujaa01/dream/synthetic.challenge.set2.normal-sorted'
````
does not work, but 

```
'samtools' 'sort' '-@ 1' '/hpc/users/ahujaa01/dream/synthetic.challenge.set2.normal.bam' '/hpc/users/ahujaa01/dream/synthetic.challenge.set2.normal-sorted'
```
does

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/16)

### DIFF
--- a/src/lib/biokepi_target_library.ml
+++ b/src/lib/biokepi_target_library.ml
@@ -226,8 +226,8 @@ module Samtools = struct
     let dest_prefix =
       sprintf "%s-%s" (Filename.chop_suffix source ".bam") "sorted" in
     let destination = sprintf "%s.%s" dest_prefix "bam" in
-    let sort_with_threads = sprintf "sort -@ %d" processors in
-    let make_command src des = [sort_with_threads; src; dest_prefix] in
+    let with_threads = sprintf "-@ %d" processors in
+    let make_command src des = ["sort"; with_threads; src; dest_prefix] in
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by smondet** [comment]
`do_on_bam` is based on `Program.exec`:

```ocaml
  let do_on_bam
      ~(run_with:Machine.t)
      ?(more_dependencies=[]) bam_file ~destination ~make_command =
    let open Ketrew.EDSL in
    let samtools = Machine.get_tool run_with "samtools" in
    let src = bam_file#product#path in
    let sub_command = make_command src destination in
    let program =
      Program.(Tool.(init samtools) && exec ("samtools" :: sub_command)) in
    let make = Machine.run_program run_with program in
    let host = Machine.(as_host run_with) in
```

The semantics of exec is to do like if there was no shell at all, the list should be a list of “`argc/argv`" to call the process.

Having `-@ 42` as one argument for samtools is a bit weird (it still works because of the way they parse the command line).

I redid the patch like that: https://github.com/hammerlab/biokepi/commit/14d65d1cbad7e21b403b3d8bae6194b083320e78

PR #17, opened 2015-04-01
Add travis config

arahuja:
I based this off of https://github.com/ocaml/ocaml-travisci-skeleton since I just wanted something to check the build, but I saw there is something more complex in ketrew.  Let me know if you think things from there are necessary

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/17)


smondet:
To use `ocaml-travis-skeleton` we also need an `opam` file in the root `biokepi` directory.

And, also, Biokepi depends on Ketrew's master branch, so we need to point to `smondet/dev-opam-repo` to get it.

Issue #30, opened 2015-04-24
Support Guacamole


smondet:
This requires a new kind of machine, that knows about Hadoop/Yarn.

I've made it work before, the question is who to abstract it to make it portable/configurable.

arahuja:
Are thinking of moving the `Spark` module into here or ketrew directly?

ryan-williams:
Hoping to breathe some life into this, I've been increasingly wanting this recently!

ihodes:
I think @e5c is beginning work on this?

smondet:
> This requires a new kind of machine, that knows about Hadoop/Yarn.

This should now be handled by the requirements thing (implementations of `Machine.t` that do not know about spark/yarn could be improved by dying with a nice error message).

e5c:
@smondet 
> This should now be handled by the requirements thing
Mind pointing to a code example where spark/yarn is specified as a requirement for a machine?

smondet:
@e5c there is no code that requires spark so far, I just meant that the "infrastructure" is there to wire the things together.

There is a `` `Spark [ ... ]`` requirement already there:
https://github.com/hammerlab/biokepi/blob/master/src/run_environment/machine.ml#L128
So any tool can call ``Machine.run_program ... ~requirements:[ `Spark ["foo"; "bar"] ]``
but it's unused so far.

If you hack on that and the `string list` argument is not the structure you find appropriate to "configure" spark/yarn usage , please change it, I just picked it semi-randomly :)

PR #39, opened 2015-05-07
Add bedtools and bamtofastq

arahuja:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/39)


hammer:
Why use `samtools` for `bam2fastq` and not ADAM?

arahuja:
I actually did use ADAM, but implemented this first so am submitting anyways.

hammer:
Hmm interesting then we'll have a naming conflict in the space of `biokepi` targets. @smondet do you have any thoughts on how targets in `biokepi` should be organized into namespaces?

smondet:
@arahuja  Thanks

@hammer it used to be complete anarchy, and @arahuja started organizing into modules, we should keep going that way.

ADAM requires more "system admin" support than Biokepi has so far, discussion there: https://github.com/hammerlab/biokepi/issues/30

PR #42, opened 2015-05-13
Bams in pipelines

smondet:
This adds “input bams” and the step “bam-to-fastq”.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/42)


arahuja:
Cool -  I was using something like `Bam_sample` locally as a way to jump straight to a variant caller - nice that is added in now.

smondet:
I used `` `Single`` and `` `Paired`` to kinda force the fact that they are different from the ones of `bamtofastq`; the translation is just one more line and (for me) makes the whole thing more readable (things that are different are called differently).
But that also may be me being over paranoid :)

PR #54, opened 2015-05-30
Name bam pair after tumor file

arahuja:
- There can be multiple somatic variant calling runs with the same normal file, but different tumor files.  Naming the run after the normal file would create a file name collision.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/54)


smondet:
I merge this as a quick fix, but we should find a longer-term solution: #56.

PR #60, opened 2015-06-15
Remove equivalence from single end

arahuja:
Not sure if you want to merge this @smondet or if you will find a better solution, but filed the PR to record the problem and possible fix.

The single-end workflow node ends up in a loop by having a dependency with the same success condition.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/60)


smondet:
So, the `~equivalence:`None` should have been the right fix, but the `` `And [r1; r1]`` hack is necessary because of https://github.com/hammerlab/ketrew/issues/189.

PR #65, opened 2015-08-13
Add md tag to star output

### LINKED ISSUE #64: add mdtag to star output (and any other aligners that don't do this already)
According to this: https://groups.google.com/forum/#!topic/rna-star/AWuKw5zuEiY there is a patch to have star output the MD field itself, but probably easier is to just run `samtools calmd` as the final step in the pipeline.

arahuja:
Fixes #64

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/65)


smondet:
The build issue is unrelated, I'll take care of Travis separately.

PR #67, opened 2015-08-24
add filter_reads_with_N_cigar to GATK HC

arahuja:
GATK HaplotypeCaller fails on DNA BAM files with CIGAR N - this flag skips those lines.  It would be nice to restrict this to BAM files that only have `~contains \`DNA` is that possible?

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/67)

### DIFF
--- a/src/lib/gatk.ml
+++ b/src/lib/gatk.ml
@@ -147,6 +147,7 @@ let haplotype_caller ~run_with ~reference_build ~input_bam ~result_prefix how  =
               "-I"; sorted_bam#product#path;
               "-R"; reference_fasta#product#path;
               "-o"; output_vcf;
+              "--filter_reads_with_N_cigar";
             ]
           )
       in


smondet:
Yes it is possible (assuming the field is correctly filled :) )

```ocaml
match input_bam#content_type with
| `DNA -> ...
| `RNA -> ...
```

c.f. https://github.com/hammerlab/biokepi/blob/master/src/lib/common.ml#L162

arahuja:
@smondet I was thinking more in the function signature?  Otherwise in the way you wrote it would just throw an exception on the `RNA` case?

smondet:
oh you want to make it impossible to call `haplotype_caller` on RNA bams?

we can add a phantom type to the bam object, but I don't think it is worth the complexity.

arahuja:
> oh you want to make it impossible to call haplotype_caller on RNA bams?

Yea that'd be ideal, since it is not really meant for that and with this option added would result in most reads being thrown out.

Issue #72, opened 2015-10-19
Generate online API documentation


rleonid:
Woops, there is a build target...

rleonid:
Done by #73

PR #73, opened 2015-10-20
Add documentation

rleonid:
I took out the call to `oredoc`, happy to add that back when `oredoc` is easier to access. This should address #72 as there is now `gh-pages` branch being [served](https://github.com/hammerlab/biokepi). But I think we should file separate issues for improving documentation.

### DIFF
--- a/_oasis
+++ b/_oasis
@@ -52,3 +52,10 @@ Executable "biokepi-demo"
   CompiledObject: best
   BuildDepends: biokepi, threads
   MainIs:     main.ml
+
+Document biokepi
+  Title:                API reference for Biokepi
+  Type:                 ocamlbuild (0.4)
+  BuildTools:           ocamlbuild, ocamldoc
+  XOCamlbuildPath:      src/lib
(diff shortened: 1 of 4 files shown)


_oasis
```diff
@@ -52,3 +52,10 @@ Executable "biokepi-demo"
   CompiledObject: best
   BuildDepends: biokepi, threads
   MainIs:     main.ml
+
+Document biokepi
+  Title:                API reference for Biokepi
+  Type:                 ocamlbuild (0.4)
+  BuildTools:           ocamlbuild, ocamldoc
+  XOCamlbuildPath:      src/lib
+  XOCamlbuildLibraries: biokepi
```
smondet: Why use Oasis?

It's missing the `-charset UTF-8 `, ` -keep-code` and `-colorize-code` options.

(And it's slower for 0 benefits).

rleonid: I switched to using Oasis because it seems consistent to keep all the capabilities in one place, and the default one that I found in the Makefile didn't work right away (so I assumed it was legacy).

PR #83, opened 2015-11-05
Add developer files

rleonid:
[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/83)


.ocamlinit
```diff
@@ -0,0 +1,3 @@
+#directory "_build/src/lib"
```
smondet: could you also `#use "topfind"` and `#thread`?

rleonid: Yep

Issue #86, opened 2015-11-11
Remove option wrapper around is_done Condition


rleonid:
Having an a product where `is_done = None` does not make sense. You should always be able to determine when a target `is_done`. We should not use it for target equivalence, since there is already a field for that logic in a `target`.

smondet:
The `equivalence` field in Ketrew targets says “how to compute equivalence”, the default value tells to look at the condition (`is_done`).

PR #87, opened 2015-11-11
No more option condition

rleonid:
This addresses #86

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/87)


smondet:
Sorry cannot merge, for ketrew `None` is different from ``Some `Satisfied`` this patch can break a lot of “phony” targets that do not set explicitly the equivalence to `` `None`` (cf. http://seb.mondet.org/software/ketrew/api/Ketrew.EDSL.html#VALtarget)

Issue #90, opened 2015-11-11
Mutect should take parameters like Strelka does


ihodes:

ihodes:
The only documentation I could find of MuTect params is here: https://github.com/broadinstitute/mutect/blob/master/src/org/broadinstitute/cga/tools/gatk/walkers/cancer/mutect/MuTectArgumentCollection.java

PR #93, opened 2015-11-13
Allow MuTect to take parameters

ihodes:
Fixes https://github.com/hammerlab/biokepi/issues/90

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/93)

### DIFF
--- a/src/lib/mutect.ml
+++ b/src/lib/mutect.ml
@@ -2,9 +2,27 @@ open Common
 open Run_environment
 open Workflow_utilities
 
+module Configuration = struct
+
+  type t = { name: string; parameters: (string * string) list; }
+
+  let to_json {name; parameters}: Yojson.Basic.json =
+    `Assoc [
+      "name", `String name;
+      "parameters",
+      `Assoc (List.map parameters ~f:(fun (a, b) -> a, `String b));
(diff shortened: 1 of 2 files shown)


src/lib/mutect.ml
```diff
@@ -2,9 +2,27 @@ open Common
...
+
+  let default =
+    {name = "default"; parameters = []}
+
+end
+
 let run ?(reference_build=`B37)
     ~(run_with:Machine.t) ~normal ~tumor ~result_prefix
-    ?(more_edges = []) how =
+    ?(more_edges = []) ~configuration how =
```
arahuja: Maybe a default argument of the default configuration?

ihodes: I do have a default in the API exposed in pipeline.ml, so in practice you don't have to specify it, but left it required here as some of the other modules do (e.g. Strelka).

smondet: He's put the optional-with-default at the `Pipeline.t` level → https://github.com/hammerlab/biokepi/pull/93/files#diff-6cc6e1c3d49a4d36b4884380ad53d585R177

One day we should make everything more homogeneous :)

Issue #85, opened 2015-11-06
Make reference-build handling more extensible


smondet:
Now we now more where we are going with these.

- remove the polymorphic-variant
- centralize the existing references into one map-like datastructure accessed by name, and extensible from outside of the library.

smondet:
We should also store sizes and md5sums of the files we want to download and check them with the `~condition` argument.

smondet:
Mostly done, comment above moved to https://github.com/hammerlab/biokepi/issues/211

PR #118, opened 2016-02-03
Revamp reference genomes

smondet:
Cf. #85.

It's not finalized but worth merging before keeping the thing going.

- Now reference genomes are accessed by names that are strings (more extensible, anyone can add reference genomes to their `Machine.t` without touching Biokepi).
- The module `Reference_genome` itself is not 2 parts; the `Specification` (that could go some time to the `redaw` library; doesn't depend on Ketrew) and the biokepi “internal representation” (that was already there).
- There is a downloading test.
- The paths to where reference genomes get downloaded have changed (if one wants to avoid redownloading/indexing, they can create a `Reference_genome.t` with local paths and add it to their `Machine.t` or, less elegant, just move the files to expected paths ☺).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/118)


src/lib/build_machine.ml
```diff
@@ -20,37 +19,22 @@ let create
...
-      | `mm10 -> Data_providers.pull_mm10
-                   ~host ~run_program ~destination_path:(meta_playground // "mm10-reference-genome")
-      )
+    ~get_reference_genome:(fun name ->
+        match name, b37 with
+        | name, Some some37 when name = Reference_genome.name some37 -> some37
+        | name, _ ->        
+          Download_reference_genomes.get_reference_genome name
+            ~toolkit ~host ~run_program
+            ~destination_path:(meta_playground // "reference-genome"))
```
ihodes: Might we want to cat the ref genome's name on to the destination?

src/lib/reference_genome.ml
```diff
@@ -1,14 +1,197 @@
 open Common
 
-type specification =
-  [`B37 | `B38 | `hg19 | `hg18 | `B37decoy | `mm10 ]
+type name = string
+
+module Specification = struct
+  module Location = struct
```
ihodes: Love this + how it's implemented.

ihodes:
LGTM, just one question. Nice change!

src/lib/build_machine.ml
smondet: The function `of_specification` actually creates a subdirectory with the genome name:
https://github.com/hammerlab/biokepi/pull/118/files#diff-443f45965cd1f3ecd1803f25ca33c469R22

And I've tested that :)
https://github.com/hammerlab/biokepi/pull/118/files#diff-cd61c399df66872fa5b8361429ff9c49R46

```
 $ ls biokepi-test-all-downloads/
b37  b37decoy  b38  hg18  hg19  mm10  tools
```

ihodes: Ah right, should've realized once I saw that test :) Awesome!

Issue #122, opened 2016-02-10
Add switch (env var like BIOKEPI_DEBUG) to switch on/off Biokepi debug statements


ihodes:
Cf. https://github.com/hammerlab/biokepi/blob/ec1f842197285d57b1a71137f4ba72f133b82d7b/src/lib/common.ml#L33

Right now, when I compile a pipeline, I get about 100 lines of debug statements. I typically don't want them, but it'd be nice to have them sometimes. This'd be an easy change due to the way they're implemented, cf. above.

Issue #116, opened 2016-01-28
Make reference annotation GTF optional for Stringtie


ihodes:
It appears to be optional; right now, an exception is thrown when there is no GTF.

Issue #108, opened 2016-01-08
Germline caller (in `pipeline.ml`) adds the `-normal` tag to any bam


smondet:
This makes the compiled pipeline do more work than necessary because of the file renaming:

https://github.com/hammerlab/biokepi/blob/25aecf5c073c5015ae0b8f20869d68a4715b010c/src/lib/pipeline.ml#L571

smondet:
This is done in #123 forgot to close the issue.

Issue #104, opened 2015-12-23
Difficult to decipher exn


ihodes:
`Exception: Failure "for now, a sample must be a list of fastq.gz".`

found https://github.com/hammerlab/biokepi/blob/master/src/lib/pipeline.ml#L111

Triggered, in my case, by not naming the file with a "fastq.gz" suffix.

PR #123, opened 2016-02-10
Small-ish various fixes

smondet:
Fixes for #104, #108, #116, #122.

@ihodes: I haven't tested Stringtie like this.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/123)

### DIFF
--- a/src/lib/common.ml
+++ b/src/lib/common.ml
@@ -30,7 +30,13 @@ end
 let (//) = Filename.concat
 (** [path // filename] will concat [filename] to the end of [path]. *)
 
-let dbg fmt = ksprintf (eprintf "biokepi-debug: %s\n%!") fmt
+let debug_mode = 
+  ref (try Sys.getenv "BIOKEPI_DEBUG" = "true" with _ -> false)
+let dbg fmt = ksprintf (fun s ->
+    if !debug_mode
+    then eprintf "biokepi-debug: %s\n%!" s
+    else ()
+  ) fmt
 (** A consistent debugging mechanism. *)
 
 let failwithf fmt = ksprintf failwith fmt
--- a/src/lib/pipeline.ml
+++ b/src/lib/pipeline.ml
@@ -83,7 +83,7 @@ type _ t =
   | Concat_text: fastq t list -> fastq t
   | Star: fastq_sample t -> bam t
   | Hisat: fastq_sample t -> bam t
-  | Stringtie: Stringtie.gtf_usage_preference * bam t -> gtf t
+  | Stringtie: Stringtie.Configuration.t * bam t -> gtf t
   | Bwa: bwa_params * fastq_sample t -> bam t
   | Bwa_mem: bwa_params * fastq_sample t -> bam t
   | Mosaik: fastq_sample t -> bam t
@@ -154,8 +154,8 @@ module Construct = struct
 
   let hisat fastq = Hisat fastq
 
-  let stringtie ?(use_reference_gtf =`If_available) bam =
-    Stringtie (use_reference_gtf, bam)
+  let stringtie ?(configuration = Stringtie.Configuration.default) bam =
+    Stringtie (configuration, bam)
 
   let gatk_indel_realigner
         ?(configuration=Gatk.Configuration.default_indel_realigner)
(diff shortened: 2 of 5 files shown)


src/lib/pipeline.ml
```diff
@@ -140,7 +154,8 @@ module Construct = struct
 
   let hisat fastq = Hisat fastq
 
-  let stringtie bam = Stringtie bam
+  let stringtie ?(use_reference_gtf =`If_available) bam =
```
iskandr: What's the alternative to `If_available` called? 

Also, what do you think of just having a boolean flag for `reference_guided_assembly` and if that's true then a GTF is required.

src/lib/pipeline.ml
```diff
@@ -316,16 +331,21 @@ let rec to_file_prefix:
       sprintf "%s-star-aligned" (to_file_prefix ?is sample)
     | Hisat (sample) ->
       sprintf "%s-hisat-aligned" (to_file_prefix ?is sample)
-    | Stringtie (sample) ->
-      sprintf "%s-stringtie" (to_file_prefix ?is sample)
+    | Stringtie (use_reference_gtf, sample) ->
+      sprintf "%s-%sstringtie"
+        (to_file_prefix ?is sample)
+        (match use_reference_gtf with
+        | `Yes -> "yesgtf"
```
iskandr: Ah, I see, it's `Yes`, `No`, and `If_available`. I don't think there's ever a case where I don't care whether I get back transcripts using known names or denovo transcripts with names like "STRG.3949". I think we should limit this trinary variable to just a regular boolean.

src/lib/pipeline.ml
smondet: It's ``[`Yes, `No, `If_available]``: https://github.com/hammerlab/biokepi/pull/123/files#diff-410ac8cc1ab382d8fa10f46e94f1fbb6R5

src/lib/pipeline.ml
smondet: The `If_available` case is not really "I don't care", I see it more as "I trust the `Reference_genome.t` (fine 90% of the time) and don't want to fail early and figure out now what I really wanted (the case that I think triggered issue #116 :))"

iskandr: Hmm, but when would we actually be indifferent to how transcripts get assembled and named? 

I guess if you're checking to see whether StringTie runs at all and don't care about the output this might be the case but I'm not sure that's common enough to add these custom not-quite-boolean options.

iskandr: It seems like Isaac's problem #116 would have been solved with just a boolean option `~use_reference_gtf:false`.

smondet: I prefer the 3-val variant; it's more readable and gives more power.

My problem is choosing the default value; I picked `` `If_available`` because it's the one *I* would prefer but maybe forcing the failure when there is no GTF is a better default (it would be a "Heads Up, something's wrong" case)

src/lib/common.ml
```diff
@@ -30,7 +30,13 @@ end
 let (//) = Filename.concat
 (** [path // filename] will concat [filename] to the end of [path]. *)
 
-let dbg fmt = ksprintf (eprintf "biokepi-debug: %s\n%!") fmt
+let debug_mode =
```
ihodes: Awesome, thank you.

src/lib/pipeline.ml
iskandr: 1) I think the only sane values here are `Yes` and `No`. The `If_available` option is just a recipe for mysterious failures downstream. 

2) ...so, since I only would want to use `Yes` and `No`, this becomes a custom boolean reimplemented with different names. Which, to me, is more confusing than just using a boolean directly. 

I'm voting for defaulting to reference guided and raising an exception early (before the job hits the Ketrew server) if the reference GTF isn't available (forcing the user to knowingly and manually select reference-free assembly).

src/lib/pipeline.ml
```diff
@@ -316,16 +331,21 @@ let rec to_file_prefix:
       sprintf "%s-star-aligned" (to_file_prefix ?is sample)
     | Hisat (sample) ->
       sprintf "%s-hisat-aligned" (to_file_prefix ?is sample)
-    | Stringtie (sample) ->
-      sprintf "%s-stringtie" (to_file_prefix ?is sample)
+    | Stringtie (use_reference_gtf, sample) ->
+      sprintf "%s-%sstringtie"
```
ihodes: I think you're missing a "-"

ihodes:
Might it be better to take the route we have with e.g. MuTect and pass in a configuration parameter (Stringtie.Configuration.t)? That way we keep doing things the same way.

I'm also unclear on why we'd use the ``If_available` option. At the very least we should have a yes/no default.

smondet:
OK switched to a `bool` inside a `Configuration.t`

ihodes:
Cool, thanks! LGTM

iskandr:
LGTM too

Issue #125, opened 2016-02-11
Add support for multi-sample indel realignment


iskandr:
When doing indel realignment on multiple samples from the same patient (e.g. normal + tumors), we almost certainly to perform indel realignment jointly on the samples (rather than independently). 

This can be accomplished by providing all the BAM files as inputs to target creator and indel realigner, and then specifying `-nWayOut realigned.bam`

For more information see:  

* http://gatkforums.broadinstitute.org/gatk/discussion/3958/indelrealigner-with-tumor-normal-pair

* http://gatkforums.broadinstitute.org/gatk/discussion/5588/best-practice-for-multi-sample-non-human-indel-realignment

arahuja:
@timodonnell opened an issue on this earlier: #41

iskandr:
Thanks @arahuja, didn't see that. 

What's the proper github etiquette, should I close this issue and copy my text over to that one?

arahuja:
Not sure there is any? Up to you I guess?

iskandr:
Closing duplicate, #41. 

Here's the text from @timodonnell: 

>My understanding of https://github.com/hammerlab/biokepi/blob/77abf040aa85cfe247a21df15e98bf68fc192839/src/lib/biokepi_common_pipelines.ml#L41 is that we are running the indel realigner separately for tumor / normal. Wenyi Wang from MD Andersen says we'll get better results if we run it jointly (which I think means just giving the gatk multiple both BAMs by specifiying the -I option twice). We should support this in biokepi (and test if we in fact get different results).

smondet:
Just pointing out that @iskandr second
[link](http://gatkforums.broadinstitute.org/gatk/discussion/5588/best-practice-for-multi-sample-non-human-indel-realignment) says that the Broad indel-realigns bams separately “in production.”

(I'm implementing this anyway; I just go lost in GATK's documentation ☺)

PR #150, opened 2016-02-25
Indel-realignment on multiple bams

smondet:
This is for #125.

The function `Gatk.indel_realigner` can take an arbitrary list of bams.

At the `Pipeline.t` level, this is, for now, implemented as an optional "compiler optimization."
It will detect pipelines of the form `somatic_variant_caller (indel |> bqsr, indel |> bqsr)` and run the indel-realigner on both the tumor and normal together and then before re-split for bqsr.

I haven't properly tested yet (I've generated the ketrew pipelines and looked at them, but haven't run GATK & co because I'm cluster-orphan these days), but while this gets run, having more eyes on the code would be nice, hence the PR :)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/150)

### DIFF
--- a/src/lib/common.ml
+++ b/src/lib/common.ml
@@ -115,6 +115,14 @@ module KEDSL = struct
         )
     end
 
+  let expolode_bam_list_node (bln : bam_list workflow_node) =
+    List.map bln#product#bams ~f:(fun bam ->
+        workflow_node bam
+          ~name:(Filename.basename bam#path)
+          ~tags:["expolode_bam_list_node"]
+          ~edges:[depends_on bln]
+          ~equivalence:`None)
+
   (* this may be overkill: *)
   type _ bam_or_bams =
     | Single_bam: bam_file workflow_node -> bam_file workflow_node bam_or_bams
--- a/src/lib/common.ml
+++ b/src/lib/common.ml
@@ -99,6 +99,27 @@ module KEDSL = struct
       method content_type = contains
     end
 
+  type bam_list = <
+    is_done:  Ketrew_pure.Target.Condition.t option;
+    bams: bam_file list;
+  >
+  let bam_list (bams : bam_file list) : bam_list =
+    object 
+      method bams = bams
+      method is_done =
+        Some (
+          `And (List.map bams
+                  ~f:(fun b ->
+                      b#is_done
+                      |> Option.value_exn ~msg:"Bams should have a Condition.t"))
+        )
+    end
+
+  (* this may be overkill: *)
+  type _ bam_or_bams =
+    | Single_bam: bam_file workflow_node -> bam_file workflow_node bam_or_bams
+    | Bam_workflow_list: bam_file workflow_node list -> bam_list workflow_node bam_or_bams
+
   let submit w = Ketrew.Client.submit_workflow w
 
 end
--- a/src/lib/common.ml
+++ b/src/lib/common.ml
@@ -115,7 +115,7 @@ module KEDSL = struct
         )
     end
 
-  let expolode_bam_list_node (bln : bam_list workflow_node) =
+  let explode_bam_list_node (bln : bam_list workflow_node) =
     List.map bln#product#bams ~f:(fun bam ->
         workflow_node bam
           ~name:(Filename.basename bam#path)
--- a/src/lib/gatk.ml
+++ b/src/lib/gatk.ml
@@ -153,7 +153,7 @@ let indel_realigner :
       (* we make this file “unique” with an MD5 sum of the input paths *)
       |> Digest.string |> Digest.to_hex in
     let output_suffix =
-      sprintf "-target-%s-%dx-%s"
+      sprintf "_target-%s-%dx-%s"
         target_config.Configuration.Indel_realigner.name
         (List.length more_input_bams + 1)
         (if more_input_bams = [] then "" else "-" ^ digest_of_input)
--- a/src/lib/gatk.ml
+++ b/src/lib/gatk.ml
@@ -139,7 +139,7 @@ let indel_realigner :
         ~f:(Samtools.sort_bam_if_necessary
               ~run_with ~processors ~by:`Coordinate) in
     let name = 
-      sprintf "gatk-indelralign-%dx-%s" 
+      sprintf "gatk-indelrealign-%dx-%s" 
         (List.length more_input_bams + 1)
         (Filename.basename input_sorted_bam_1#product#path) in
     let gatk = Machine.get_tool run_with Tool.Default.gatk in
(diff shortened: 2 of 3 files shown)


src/lib/common.ml
```diff
@@ -99,6 +99,35 @@ module KEDSL = struct
...
+      method is_done =
+        Some (
+          `And (List.map bams
+                  ~f:(fun b ->
+                      b#is_done
+                      |> Option.value_exn ~msg:"Bams should have a Condition.t"))
+        )
+    end
+
+  let expolode_bam_list_node (bln : bam_list workflow_node) =
```
ihodes: explode*

src/lib/gatk.ml
```diff
@@ -93,67 +93,149 @@ module Configuration = struct
...
+     See also 
+     http://gatkforums.broadinstitute.org/gatk/discussion/5588/best-practice-for-multi-sample-non-human-indel-realignment
+
+     On top of that we the GADT `_ KEDSL.bam_or_bams` to return have 2 possible
+     return types:
+     bam_file workflow_node or bam_list workflow_node
+  *)
+open Configuration
+let indel_realigner :
+  type a.
```
ihodes: Forgive my slowness; why is this existential type needed? & Why is it erased in the GADT above?

src/lib/pipeline.ml
```diff
@@ -587,12 +593,47 @@ module Compiler = struct
...
       match pipeline with
+      | Somatic_variant_caller
+          (som_vc,
+           Bam_pair (
+             Gatk_bqsr (n_bqsr_config, Gatk_indel_realigner (n_gir_conf, n_bam))
+             , 
+             Gatk_bqsr (t_bqsr_config, Gatk_indel_realigner (t_gir_conf, t_bam))
+           )) 
+        when has_option compiler `Multi_sample_indel_realignment
+          && n_gir_conf = t_gir_conf ->
```
ihodes: might be good to throw if the configurations are different but `Multi_sample_indel_realignment is specified; that seems like a clear user error

src/lib/gatk.ml
```diff
@@ -93,67 +93,149 @@ module Configuration = struct
...
+    let indel_config, target_config = configuration in
+    let input_sorted_bam_1 =
+      Samtools.sort_bam_if_necessary
+        ~run_with ~processors ~by:`Coordinate input_bam_1 in
+    let more_input_sorted_bams =
+      List.map more_input_bams
+        ~f:(Samtools.sort_bam_if_necessary
+              ~run_with ~processors ~by:`Coordinate) in
+    let name = 
+      sprintf "gatk-indelralign-%dx-%s"
```
ihodes: indelrealign*

src/lib/gatk.ml
```diff
@@ -93,67 +93,149 @@ module Configuration = struct
...
+      | Single_bam b ->
+        (* This is what we give to the `-o` option: *)
+        workflow_node  ~name ~make ~edges
+          (bam_file ~sorting:`Coordinate ~host:Machine.(as_host run_with)
+             (output_bam_path b))
+      | Bam_workflow_list bamlist ->
+        workflow_node  ~name ~make ~edges
+          (bam_list 
+             (List.map bamlist ~f:(fun b ->
+                  (* This is what the documentation says it will to
```
ihodes: :pray: we hope!

(+ will do*)

ihodes:
LGTM % couple comments! Thanks for handling this, bit more complicated than I'd have guessed.

src/lib/gatk.ml
smondet: This is not an existential; this to force the “for all `a`, if my input is `a` bam_or_bams, return `a`.”

src/lib/pipeline.ml
smondet: Now the option has both options (pass silently or fail if it does not happen)

smondet:
@ihodes It looks more complicated because of the geeky types involved in reusing the same code as much as possible (creating a `indel_realignment_multi` function would have been much easier but there would have been copy-pasted code)

ihodes:
LGTM!

src/lib/pipeline.ml
```diff
@@ -587,12 +594,57 @@ module Compiler = struct
...
+                              Bam_sample (realigned_tumor#product#path,
+                                          realigned_tumor)))))
+          in
+          compile_variant_caller_step ~compiler new_pipeline
+        | other ->
+          failwithf "Gatk.indel_realigner did not return the correct list \
+                     of length 2 (tumor, normal): it gave %d bams"
+            (List.length other)
+        end
+      | Somatic_variant_caller (som_vc, Bam_pair (_, _)) as soma_pipeline
```
smondet: This is too strong of a condition; the `new_pipeline` created just above already breaks it :)

Fixing this.

src/lib/gatk.ml
```diff
@@ -93,67 +93,149 @@ module Configuration = struct
...
+        ] @ Indel_realigner.render indel_config @
+        begin match more_input_bams with
+        | [] ->
+          ["-o";
+           Filename.chop_extension
+             input_sorted_bam_1#product#path ^ output_suffix ^ ".bam" ;]
+        | more ->
+          List.concat_map more
+            ~f:(fun b -> ["-I"; Filename.quote b#product#path])
+          @ ["--nWayOut"; output_suffix ^ ".bam"]
```
smondet: Think we need to escape the `output_suffix` or at least not start with `-`:

```
##### ERROR ------------------------------------------------------------------------------------------
##### ERROR A USER ERROR has occurred (version 3.5-0-g36282e4): 
##### ERROR
##### ERROR This means that one or more arguments or inputs in your command are incorrect.
##### ERROR The error message below tells you what is the problem.
##### ERROR
##### ERROR If the problem is an invalid argument, please check the online documentation guide
##### ERROR (or rerun your command with --help) to view allowable command-line arguments for this tool.
##### ERROR
##### ERROR Visit our website and forum for extensive documentation and answers to 
##### ERROR commonly asked questions http://www.broadinstitute.org/gatk
##### ERROR
##### ERROR Please do NOT post this error to the GATK forum unless you have really tried to fix it yourself.
##### ERROR
##### ERROR MESSAGE: Argument with name 'target-ignore-mismatch-2x--b7beee7a8d8b58aac8b3d9321229954e.bam' isn't defined.
##### ERROR ------------------------------------------------------------------------------------------
```

But the call was actually `` --nWayOut -target-ignore-mismatch-2x--b7beee7a8d8b58aac8b3d9321229954e.bam`` (with one `-` that's not in the error message)

smondet:
I fixed the issues above and it seems to have run successfully.

ihodes:
LGTM! Thanks for the feature!

Issue #135, opened 2016-02-16
Specify INDEL and SNP dbSNP VCFs separately


ihodes:
cf. https://github.com/hammerlab/biokepi/issues/134

Issue #134, opened 2016-02-16
Fix mm10's dbSNP VCF(s)


ihodes:
They shouldn't be merged; BQSR only cares about the SNPs. 

They should be sorted (this caused the BQSR error).

This leads to a bigger problem; we should be able to specify & manipulate the INDEL and SNP dbSNP VCFs separately, cf.  https://github.com/hammerlab/biokepi/issues/135

PR #152, opened 2016-03-01
Add sorting of DbSNP VCFs

smondet:
This adds a `vcf-sort` step after downloading MM10's DbSNP's SNPs and INDELs.

So this is a fix for #134 to run BQSR on MM10, but not for #135 for which we need
to come up with decent semantics for `#dbsnp` and a consistent API.

I've run
[`src/test/all_downloads.ml`](https://github.com/hammerlab/biokepi/blob/master/src/test/all_downloads.ml)
and the result looks sorted.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/152)

### DIFF
--- a/src/lib/vcftools.ml
+++ b/src/lib/vcftools.ml
@@ -4,6 +4,10 @@ open Workflow_utilities
 
 
 
+(** 
+   Call a command on a list of [~vcfs] to produce a given [~final_vcf] (hence
+   the {i n-to-1} naming).
+*)
 let vcf_process_n_to_1_no_machine
     ~host
     ~vcftools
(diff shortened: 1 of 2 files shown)


smondet:
```
$ cat dbsnp.vcf | grep -in '^X' | awk '{ print $1 }' > lines_of_x.txt
$ sort -u lines_of_x.txt > lines_of_x-sorted.txt
$ diff lines_of_x.txt lines_of_x-sorted.txt && echo The Same || echo Not The Same
The Same
```

src/lib/vcftools.ml
```diff
@@ -3,38 +3,60 @@ open Run_environment
 open Workflow_utilities
 
 
+
+let vcf_process_n_to_1_no_machine
```
ihodes: What does this mean?

ihodes:
LGTM, just not sure about the naming/purpose of the vcf_process_n_to_1_no_machine. Thanks for the fix!

PR #161, opened 2016-03-09
Do various fixes

smondet:
Now that I have a cluster, could do more tests:

- GATK's indel-realigner with `--nWayOut` also does a `basename` on the input bams so it outputs to the “current” directory. Hence had to `cd` there to have proper file paths.
- There was a bug in `Pipeline.to_file_prefix` that the optmization pass exhibited.
- There was a 99.9% silent bug in `Gatk.bqsr` with the path to the `-recal_data.table` file that I saw while reading the code.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/161)


ihodes:
LGTM! Thanks for the fixes.

Issue #157, opened 2016-03-02
Parallelize BWA across FASTQs 


iskandr:
The alignments produced by BWA are independent across reads. 

Each FASTQ can be aligned separately, for example:

```
# align all tumor FASTQs
bwa aln tumor_L005_001_R1.fastq.gz > tumor_L005_001_R1.sai
bwa aln tumor_L005_001_R2.fastq.gz > tumor_L005_001_R2.sai
bwa aln tumor_L005_002_R1.fastq.gz > tumor_L005_002_R1.sai
bwa aln tumor_L005_002_R2.fastq.gz > tumor_L005_002_R2.sai
# align normal FASTQs
bwa aln normal_L004_001_R1.fastq.gz > normal_L004_001_R1.sai
bwa aln normal_L004_001_R2.fastq.gz > normal_L004_001_R2.sai
```

(see: https://wikis.utexas.edu/display/CoreNGSTools/Alignment#Alignment-Performingthebwaalignment)

Once we have a collection of aligned `sai` files, each R1/R2 `sai` pair can be converted to a SAM file:

```
bwa sampe grch38.fa tumor_L005_001_R1.sai tumor_L005_001_R2.sai tumor_L005_001_R1.fastq tumor_L005_001_R2.fastq > tumor_L005_001.sam
```

The `sam` files for each R1/R2 pair can be converted to `bam` files (with `samtools view`) and then merged into a single `bam` file with e.g. `samtools merge tumor.bam tumor*.bam`

I'm surprised that there isn't much documentation online about doing this but it does seem that aside from combining mate pairs, nothing done by `bwa` involves interactions between reads. 

A similar pipeline seems to be implemented in Snakemake: https://github.com/inodb/snakemake-parallel-bwa/blob/master/Snakefile

iskandr:
Hopefully something similar can be done for MOSAIK, but making sure separate alignment of FASTQs won't give different results: https://github.com/wanpinglee/MOSAIK/issues/24

iskandr:
@smondet I think I'd rather create the workflow nodes for parallelization explicitly than rely on a compiler optimization (unless there's some way to enforce that this optimization has to run). I think that for larger sequencing runs it can't be thought of as optional (especially when we're under a tight time budget)

smondet:
@iskandr what I'm doing there is adding the option to fail loudly if the compiler pass does not happen:
[...]

PR #162, opened 2016-03-09
Optimize aligners over FASTQ fragments on-demand

smondet:
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/162)

### DIFF
--- a/src/lib/pipeline.ml
+++ b/src/lib/pipeline.ml
@@ -463,8 +463,12 @@ let rec to_json: type a. a t -> json =
 
 module Compiler = struct
   type 'a pipeline = 'a t
+  type workflow_option_failure_mode = [ `Silent | `Fail_if_not_happening ] 
   type workflow_option = [
-    | `Multi_sample_indel_realignment of [ `Silent | `Fail_if_not_happening ]
+    | `Multi_sample_indel_realignment of workflow_option_failure_mode
+    | `Parallel_alignment_over_fastq_fragments of
+        [ `Bwa_mem | `Bwa | `Mosaik | `Star | `Hisat ] list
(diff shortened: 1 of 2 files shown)


smondet:
It fixes https://github.com/hammerlab/biokepi/issues/157

src/lib/pipeline.ml
```diff
@@ -545,6 +549,67 @@ module Compiler = struct
...
+        | None -> `Not_caught
+        end
+      | Paired_end_sample (name,
+                           Gunzip_concat r1_list,
+                           Gunzip_concat r2_list)
+        when List.length r1_list = 1 -> `Not_caught
+      | other ->
+        begin match option with
+        | Some `Silent
+        | None -> `Not_caught
```
ihodes: I think the default of when a compiler optimization cannot be made, but is expected to by the user, should be noisy alert, not silent failure and continuing with the compilation.

ihodes: Ah I see; this is when there is no optimization specified. I recall my previous comment.

src/lib/pipeline.ml
```diff
@@ -545,6 +549,67 @@ module Compiler = struct
           ~name:(sprintf "single-end %s"
                    (Filename.basename r1#product#path))
     in
+    let parallelize_alignment ~make_aligner sample =
+      match sample with
+      | Paired_end_sample (name,
+                           Gunzip_concat r1_list,
+                           Gunzip_concat r2_list) ->
+        let exploded =
+          let count = ref 0 in
```
ihodes: ლ(ಥ Д ಥ )ლ

ihodes:
LGTM!

PR #164, opened 2016-03-11
Adding opam

rleonid:
I'm opening this PR early to solicit feedback on whether this is the right style and organization of biokepi.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/164)


smondet:
Why is there a `src/lib/foo`?

I'm not sure i see the point of the `Biokepi.Opam` module, what do you plan to put there? opam update/upgrade?

src/lib/run_environment.ml
```diff
@@ -28,10 +28,11 @@ module Tool = struct
       [@@@ocaml.warning "-11"]
       type t = [
         | `Bwa of [ `V_0_7_10 ]
-        | `Custom of string * string
+        (* A tool that is manually installed by retrieving source or binary.*)
```
smondet: Why “manually”? most tools defined as `` `Custom`` are installed by biokepi.

src/lib/opam.ml
```diff
@@ -0,0 +1,12 @@
...
+
+let version ~(run_with:Machine.t) =
+  let open KEDSL in
+  let opam_tool = Machine.get_tool run_with Tool.Default.opam in
+  let name = sprintf "opam version" in
+  let make =
+    Machine.run_program run_with
+      Program.(Tool.init opam_tool && sh "opam --version")
+  in
+  workflow_node ~name ~make without_product
```
smondet: A workflow node using a tool should depend on its `ensure` node.

src/lib/run_environment.ml
rleonid: I wasn't certain of this. `opam` as a tool doesn't seem that much different than all of the other stuff defined as `custom`. I couldn't figure out what made `bwa` special.

rleonid: Oh, you're commenting on the documentation! Yes, now I do find it confusing as well. So what's the reason for the ADT in these definitions?

smondet: `` `Bwa _`` was the first one there, meant as an example. Many things in `` `Custom`` were meant to be upgraded to their own variant. But this never seemed a priority.

rleonid:
the `foo` is gone. Yes, I think we would use the `update` and `install` commands to get other tools via `biopam`. I think the next command would be to `opam init -n --compiler=0.0.0 biopam https://github.com/solvuu/biopam`

smondet:
OK then I think the `Biokepi.Opam` module should be named `Opam_tool_provider` (or alike) and it cannot use `Machine.t` because it will be used to create `Machine.t` values (circular dependency).

(That or we revamp the `Machine` module for the whole library; we can break the circular dependency with less purity maybe)

src/lib/run_environment.ml
rleonid: So it is for the convenience of `Definition.t`. Are there other parameters (fields) that a tool needs? Perhaps changing it to a simpler record

rleonid:
I see, and then modify `Build_machine.create` so that we can create a `Machine.t` from `biopam` ?

smondet:
`Build_machine.create` already uses the default `Tool.Kit.t` so it gets `opam` or any tool that uses biopam and we put there for free.

Also `Build_machine.create` is nice for quick testing but it's not intendend to be use "large scale" (most clusters will need to got back to `Machine.create` to set parameters more *fine-grainly*).

Not sure I'm answering your question :) (?)

rleonid:
That makes sense. I think my confusion stems from a lack of a public facing Biokepi API, but that's it is still a WIP so that is to expected.

src/lib/tool_providers.ml
```diff
@@ -468,13 +461,33 @@ let muse_tool ~host ~meta_playground =
...
   Tool.create Tool.Default.muse ~ensure ~init
 
+let opam_tool ~host ~meta_playground =
+  let open KEDSL in
+  let url =
+    "https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-Linux"
+  in
+  let install_path = meta_playground // "opam" in
+  let binary_path = install_path // "opam" in
+  let ensure =
```
smondet: why not use `Biopam.installed`?

src/lib/biopam.ml
```diff
@@ -0,0 +1,93 @@
...
+       the rest of the arguments. *)
+    if switch then
+      Printf.sprintf ("OPAMROOT=%s OPAMSWITCH=0.0.0 %s " ^^ fmt) root bin
+    else
+      Printf.sprintf ("OPAMROOT=%s %s " ^^ fmt) root bin
+
+  let file ?(switch=true) ~meta_playground ~package f =
+    let s = com ~switch ~meta_playground "config var %s:lib" package in
+    (* Are there places where this tick logic is flaky? *)
+    (Printf.sprintf "`%s`" s) // f
```
smondet: Maybe this should be called "opam_file_command" ?

Something called `file` will make us use `Filename.quote` which would break the backquotes thing.

src/lib/tool_providers.ml
rleonid: Ah, good point, this was from the first pass. I'll get rid of it as a tool in `Tool_providers` entirely.

src/lib/biopam.ml
rleonid: good point, how about `file_command` since we're already in the `Opam` module?

PR #169, opened 2016-03-14
Add Picard.sort_vcf

ihodes:
This is necessary for some Picard tools requiring contigs to appear in
the same order in both the reference and a VCF like dbSNP.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/169)

### DIFF
--- a/src/lib/picard.ml
+++ b/src/lib/picard.ml
@@ -24,6 +24,48 @@ let create_dict ~(run_with:Machine.t) fasta =
       on_failure_activate (Remove.file ~run_with dest);
     ]
 
+(* Sort the given VCF.
+
+- [?sequence_dict] A sequence dictionary by which the VCF will be sorted.
+*)
+let sort_vcf ~(run_with:Machine.t) ?(sequence_dict) input_vcf =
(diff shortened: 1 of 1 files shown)


src/lib/picard.ml
```diff
@@ -24,6 +24,48 @@ let create_dict ~(run_with:Machine.t) fasta =
...
+    | Some d -> sprintf "SEQUENCE_DICTIONARY= %s" (Filename.quote d#product#path)
+  in
+  let sequence_dict_edge =
+    match sequence_dict with
+    | None -> []
+    | Some d -> [depends_on d]
+  in
+  let program =
+    Program.(Tool.(init picard) &&
+             (shf "java -jar $PICARD_JAR CreateSequenceDictionary %s I= %s O= %s"
```
smondet: Should the command be `SortVcf`?
https://broadinstitute.github.io/picard/command-line-overview.html#SortVcf

ihodes: Ah yes, thank you! Fixed.

PR #170, opened 2016-03-15
Fix missing `~host` in `get_broad_jar`

smondet:
It seems that since the last time we changed all that code, we have tested this
one only with Ketrew servers that have a shared file-system with the host that
runs the jobs.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/170)


iskandr:
LGTM

ihodes:
Nice catch, thanks! (LGTM2)

PR #173, opened 2016-03-16
Conda and seq2HLA as demo

rleonid:
Exposing `conda` very similarly to the way we expose `opam`.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/173)

### DIFF
--- a/src/lib/conda.ml
+++ b/src/lib/conda.ml
@@ -1,0 +1,123 @@
+(*
+  Conda is a Python environment and package manager:
+  http://conda.pydata.org/docs/
+
+  We use it to ensure a consistent Python environment for tools that depend
+  on Python.
+*)
+
+open Common
+open Run_environment
+
+let rm_path = Workflow_utilities.Remove.path_on_host
+
+let dir ~meta_playground = meta_playground // "conda_dir"
+let commands ~meta_playground com = dir ~meta_playground // "bin" // com
+let bin = commands "conda"
+let activate = commands "activate"
+let deactivate = commands "deactivate"
+
+(* give a conda command. *)
+let com ~meta_playground fmt =
+  Printf.sprintf ("%s " ^^ fmt) (bin ~meta_playground)
+
+(* A workflow to ensure that conda is installed. *)
+let installed ~host ~meta_playground =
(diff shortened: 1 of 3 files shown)


src/lib/conda.ml
```diff
@@ -0,0 +1,123 @@
...
+
+(* give a conda command. *)
+let com ~meta_playground fmt =
+  Printf.sprintf ("%s " ^^ fmt) (bin ~meta_playground)
+
+(* A workflow to ensure that conda is installed. *)
+let installed ~host ~meta_playground =
+  let open KEDSL in
+  let url =
+    "https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh"
```
arahuja: Does 2 vs 3 matter here?

arahuja:
@leo is the assumption here a single conda environment for biokepi or per tool?

src/lib/conda.ml
rleonid: Not really, only that 3 is more "forward" looking. The environment that is defined/configured below actually uses `Python2.7`, with the idea being that eventually we'll be able parameterize by stable environment.

rleonid:
The current design is for `biokepi` but, similar to how the configuration is for just `Python2.7` now we could have a configuration for each tool.

src/lib/conda.ml
```diff
@@ -0,0 +1,123 @@
...
+let env_name = "biokepi"
+
+(* Ensure that a file exists describing the default linux conda enviroment.
+   TODO: figure out a better solution to distribute this configuration.  *)
+let cfg_exists ~host ~meta_playground =
+  let open KEDSL in
+  let file = meta_playground // config in
+  let make = daemonize ~host
+      Program.(shf "echo '%s' >> %s"
+{python|# This file may be used to create an environment using:
```
smondet: `"%s" (Filename.quote {python|....` would be more future-proof than `"'%s'" ...` because it's easy to add single-quotes.

Also why `{python| |python}`, is this python syntax?

rleonid: Addressed in latest commit.

PR #178, opened 2016-03-17
Use a GADT to model the germline & somatic variant calling distinction

rleonid:
a WIP.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/178)

### DIFF
--- a/src/lib/pipeline.ml
+++ b/src/lib/pipeline.ml
@@ -36,33 +36,24 @@ type bwa_params = {
   gap_open_penalty: int;
   gap_extension_penalty: int;
 }
-module Somatic_variant_caller = struct
-  type t = {
-    name: string;
-    configuration_json: json;
-    configuration_name: string;
-    make_target:
-      reference_build: Reference_genome.name ->
-      run_with:Run_environment.Machine.t ->
-      normal: bam ->
-      tumor: bam ->
-      result_prefix: string ->
-      processors: int ->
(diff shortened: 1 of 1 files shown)


iskandr:
I don't know if the old code handled this, but I just wanted to point out that Varscan2 has distinct modes for [single-sample variant calling](http://varscan.sourceforge.net/using-varscan.html#v2.3_pileup2snp),  [multi-sample variant calling](http://varscan.sourceforge.net/using-varscan.html#v2.3_mpileup2snp) and [somatic tumor/normal variant calling](http://varscan.sourceforge.net/using-varscan.html#v2.3_somatic). 

I'm not sure if these should be treated as different tools in Biokepi.

iskandr:
Just in case that's not enough, Varscan2 also calls copy numbers (http://varscan.sourceforge.net/using-varscan.html#v2.3_copynumber), though apparently only for tumor/normal somatic samples.

rleonid:
@iskandr Perhaps filing an issue so that we don't forget. Adding new types of variant calling is outside the scope of this line of work; which is to further understanding by simplifying the code.

smondet:
@iskandr that's why the functions are called `Varscan.somatic_*`, it's the only case implemented so far but we can add the other ones
https://github.com/hammerlab/biokepi/blob/master/src/lib/varscan.ml#L104

Issue #179, opened 2016-03-17
{HISAT,Kallisto} index ~processors


smondet:
It seems that we don't use the `processors` value in the command itself:
https://github.com/hammerlab/biokepi/blob/2aa2905b575db1188735493daad2aa0b11779503/src/lib/hisat.ml#L29

smondet:
Idem for Kallisto: https://github.com/hammerlab/biokepi/blob/2aa2905b575db1188735493daad2aa0b11779503/src/lib/kallisto.ml#L28-L34

arahuja:
Neither HISAT or Kallisto's index/build steps accept any threading arguments - should the argument be dropped then?

```
HISAT version 0.1.6-beta by Daehwan Kim (infphilo@gmail.com, http://www.ccb.jhu.edu/people/infphilo)
Usage: hisat2-build [options]* <reference_in> <bt2_index_base>
    reference_in            comma-separated list of files with ref sequences
    hisat_index_base          write bt2 data to files with this dir/basename
Options:
    -c                      reference sequences given on cmd line (as
                            <reference_in>)
    --large-index           force generated index to be 'large', even if ref
                            has fewer than 4 billion nucleotides
    -a/--noauto             disable automatic -p/--bmax/--dcv memory-fitting
    -p/--packed             use packed strings internally; slower, uses less mem
    --bmax <int>            max bucket sz for blockwise suffix-array builder
    --bmaxdivn <int>        max bucket sz as divisor of ref len (default: 4)
    --dcv <int>             diff-cover period for blockwise (default: 1024)
    --nodc                  disable diff-cover (algorithm becomes quadratic)
    -r/--noref              don't build .3/.4.bt2 (packed reference) portion
    -3/--justref            just build .3/.4.bt2 (packed reference) portion
    -o/--offrate <int>      SA is sampled every 2^offRate BWT chars (default: 5)
    -t/--ftabchars <int>    # of chars consumed in initial lookup (default: 10)
    --localoffrate <int>    SA (local) is sampled every 2^offRate BWT chars (default: 3)
    --localftabchars <int>  # of chars consumed in initial lookup in a local index (default: 6)
    --seed <int>            seed for random number generator
    -q/--quiet              verbose output (for debugging)
    -h/--help               print detailed description of tool and its options
    --usage                 print this usage message
    --version               print version information and quit
```
```
kallisto 0.42.3
Builds a kallisto index

Usage: kallisto index [arguments] FASTA-files

Required argument:
[...]

Issue #166, opened 2016-03-11
Tools should be able to estimate required memory for backends which require it


ihodes:
This makes scheduling faster, and ensures jobs aren't needlessly killed by the scheduler (here's looking at you, YARN) when they go over their requested amount.

iskandr:
This seems to usually be a function of (some subset of) reference size, number of reads, and degree of parallelism.

ihodes:
Running into something which could benefit from this:

```
Mar 22 07:03:21 ..... Started sorting BAM
EXITING because of fatal ERROR: not enough memory for BAM sorting: 
SOLUTION: re-run STAR with at least --limitBAMsortRAM 65209053098
Mar 22 07:03:21 ...... FATAL ERROR, exiting
```

in STAR alignment

Issue #163, opened 2016-03-10
Quick_run_program vs. Run_program


ihodes:
When we use `Quick_run_program` internally, we use it with the implicit assumption that commands it executes have internet access, whereas `Run_program` does not. This is because our compute nodes don't have internet access (they shouldn't). However, this is really overloading the API (`Quick_run_program` is about running short-lived commands that don't require much config, not about internet access), and manifests in Biokepi by virtue of the fact that all tool installers from `Tool_providers` use `Quick_run_program`. 

We should encode the requirement of internet connectivity in the type system, and let `Quick_run_program` and `Run_program` determine how to dispatch commands which require the internet.

smondet:
Yes! We could even merge the two functions and let the implementation decide what to do depending on the requirements specified in the arguments.

ihodes:
That would be awesome, highly in favor of that option.

ihodes:
Tools should be able to request internet access, a YARN cluster (that can e.g. run a Spark job), maybe GPUs, minimum memory requirements, etc.

iskandr:
:+1: To merging `Run_program` and `Quick_run_program`

ihodes:
This has been fixed in #180

PR #180, opened 2016-03-17
Implement “requirements” for `Machine.run_program`

smondet:
The idea is that users of the `Make_fun.t` functions express their needs with a
`~requirements` argument.

Both the `Make_fun` and `Machine` modules provide functions with default
values.

The `Make_fun.Requirement.t` variants are of course just a first stab to be
refined later (and some cases are not used anywhere).

This “fixes” #163.

We can start taking notes when we run tools, so that they can give a better
estimate of their memory requirements (cf. #166).

I came across #179 but did not fix it yet (I actually don't really know these
tools).

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/180)

### DIFF
--- a/src/lib/run_environment.ml
+++ b/src/lib/run_environment.ml
@@ -123,9 +123,6 @@ end
 
 module Machine = struct
 
-  (* type run_function = ?name:string -> ?processors:int -> Program.t -> *)
-  (*   Ketrew_pure.Target.Build_process.t *)
-
   type t = {
     name: string;
     ssh_name: string;
--- a/src/lib/run_environment.ml
+++ b/src/lib/run_environment.ml
@@ -112,9 +112,10 @@ module Make_fun = struct
     Program.t ->
     KEDSL.Build_process.t
 
-  let streamish requirements = `Processors 1 :: `Memory `Decent :: requirements
+  let stream_processor requirements = `Processors 1 :: `Memory `Decent :: requirements
   let quick requirements = `Quick_run :: requirements
-  let downloading requirements = `Internet_access :: streamish requirements 
+  let downloading requirements =
+    `Internet_access :: stream_processor requirements 
 
   let with_requirements : t -> Requirement.t list -> t = fun f l ->
     fun ?name ?(requirements = []) prog ->
@@ -149,17 +150,19 @@ module Machine = struct
   let quick_run_program t : Make_fun.t =
     Make_fun.with_requirements t.run_program (Make_fun.quick [])
 
-  (** Run a program that does not use much memory and runs on one core *)
-  let run_streamish_program t : Make_fun.t =
-    Make_fun.with_requirements t.run_program (Make_fun.streamish [])
+  (** Run a program that does not use much memory and runs on one core. *)
+  let run_stream_processor t : Make_fun.t =
+    Make_fun.with_requirements t.run_program (Make_fun.stream_processor [])
 
-  (** Run a program that does not use much memory, runs on one core, and uses the internet *)
+  (** Run a program that does not use much memory, runs on one core, and needs
+      the internet. *)
   let run_download_program t : Make_fun.t =
     Make_fun.with_requirements t.run_program (Make_fun.downloading [])
 
   let run_big_program t : ?processors: int -> Make_fun.t =
     fun ?(processors = 1) ->
-      Make_fun.with_requirements t.run_program [`Memory `Big; `Processors processors]
+      Make_fun.with_requirements
+        t.run_program [`Memory `Big; `Processors processors]
 
   let work_dir t = t.work_dir
 
--- a/src/lib/run_environment.ml
+++ b/src/lib/run_environment.ml
@@ -94,10 +94,37 @@ module Tool = struct
   end
 end
 
+module Make_fun = struct
+  module Requirement = struct
+    type t = [
+      | `Processors of int
+      | `Internet_access
+      | `Memory of [ `GB of float | `Decent | `Big ]
+      | `Quick_run
+      | `Spark of string list (* custom parameters *)
+      | `Custom of string
+    ] [@@deriving yojson, show]
+  end
+
+  type t =
+    ?name: string ->
+    ?requirements: Requirement.t list ->
+    Program.t ->
+    KEDSL.Build_process.t
+
+  let streamish requirements = `Processors 1 :: `Memory `Decent :: requirements
+  let quick requirements = `Quick_run :: requirements
+  let downloading requirements = `Internet_access :: streamish requirements 
+
+  let with_requirements : t -> Requirement.t list -> t = fun f l ->
+    fun ?name ?(requirements = []) prog ->
+      f ?name ~requirements:(l @ requirements) prog
+end
+
 module Machine = struct
 
-  type run_function = ?name:string -> ?processors:int -> Program.t ->
-    Ketrew_pure.Target.Build_process.t
+  (* type run_function = ?name:string -> ?processors:int -> Program.t -> *)
+  (*   Ketrew_pure.Target.Build_process.t *)
 
   type t = {
     name: string;
@@ -105,15 +132,14 @@ module Machine = struct
     host: Host.t;
     get_reference_genome: string -> Reference_genome.t;
     toolkit: Tool.Kit.t;
-    quick_command: Program.t -> Ketrew_pure.Target.Build_process.t;
-    run_program: run_function;
+    run_program: Make_fun.t;
     work_dir: string;
   }
   let create
       ~ssh_name ~host ~get_reference_genome ~toolkit
-      ~quick_command ~run_program ~work_dir  name =
+      ~run_program ~work_dir  name =
     {name; ssh_name; toolkit; get_reference_genome; host;
-     quick_command; run_program; work_dir}
+     run_program; work_dir}
 
   let name t = t.name
   let ssh_name t = t.ssh_name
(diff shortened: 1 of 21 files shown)


src/lib/run_environment.ml
```diff
@@ -94,10 +94,37 @@ module Tool = struct
...
+  module Requirement = struct
+    type t = [
+      | `Processors of int
+      | `Internet_access
+      | `Memory of [ `GB of float | `Decent | `Big ]
+      | `Quick_run
+      | `Spark of string list (* custom parameters *)
+      | `Custom of string
+    ] [@@deriving yojson, show]
+  end
```
rleonid: Why not use a regular variant?

src/lib/run_environment.ml
```diff
@@ -94,10 +94,37 @@ module Tool = struct
   end
 end
 
+module Make_fun = struct
+  module Requirement = struct
+    type t = [
+      | `Processors of int
+      | `Internet_access
+      | `Memory of [ `GB of float | `Decent | `Big ]
```
rleonid: Having two ambiguous options like this is just asking for trouble. Both are subjective and uninformative.

src/lib/run_environment.ml
smondet: 1. It's exported; users of the library have to match on it.
2. Also within the library, we have to put/compute actual literals
3. Polymorphic variants allow nice nesting (like `` | `Memory of [ `GB of float | `Decent | `Big ]``
4. No good reason to use regular variants (the performance difference is meaningless)

rleonid:
Where are these requirements enforced?

src/lib/run_environment.ml
rleonid: 1. You can export a regular variant, same match requirement.
2. I don't understand what that means.
3. And long nested nested statements that are hard to decipher.
4. Simplicity is a good reason.

src/lib/run_environment.ml
smondet: yes they are “subjective and uninformative" but that's all we know so far.

That's why I said it's a first stab; all I “know” is that some jobs are “normal” (= decent) and some others are "big" (will fail with less than, say, 4 GB allocated to them and their size increases with data size).

"quick-run" means "if possible I'd prefer to be run on the login node of a cluster because the scheduling part is slower than the actual job"

src/lib/run_environment.ml
```diff
@@ -94,10 +94,37 @@ module Tool = struct
...
+  let with_requirements : t -> Requirement.t list -> t = fun f l ->
+    fun ?name ?(requirements = []) prog ->
+      f ?name ~requirements:(l @ requirements) prog
+end
+
 module Machine = struct
 
-  type run_function = ?name:string -> ?processors:int -> Program.t ->
-    Ketrew_pure.Target.Build_process.t
+  (* type run_function = ?name:string -> ?processors:int -> Program.t -> *)
```
ihodes: This can be dropped.

smondet:
@rleonid users of the library, when they create `Machine.t`s they choose to enforce/ignore the requirements. cf. https://github.com/hammerlab/biokepi/issues/163

For example in `Build_machine` (= creating dummy/example `Machine.t`s) we don't know anything about the computing environment so we just ignore all requirements: https://github.com/hammerlab/biokepi/pull/180/files#diff-cd61c399df66872fa5b8361429ff9c49R21

src/lib/run_environment.ml
ihodes: I'm as much a fan of regular variants as the next guy, but keeping this open means consumers of Biokepi can add their own tools with their own requirements without having to modify Biokepi, right?

rleonid: They can, but then it won't be a `Requirement.t`, specifically one won't be able to pass them to `with_requirements`

ihodes: Given that, what's the benefit of using regular variants, given that we'd like to keep Biokepi (easily) extensible?

src/lib/run_environment.ml
ihodes: I also don't know what \`Decent or \`Big mean; perhaps we can leave it at \`Gb for now? And if gets too cumbersome, we can reconsider?

rleonid: > all I “know” is that some jobs are “normal” (= decent) and some others are "big" (will fail with less than, say, 4 GB allocated to them and their size increases with data size).

Why not write those, or better yet, if you don't know _don't_ add unneeded complexity until we need it. As a user, unless you're sitting next to me or on slack, how do I know whether `rsync` is `Decent` or `Big`. Why add an extra option just so that a user can make a mistake?

rleonid: `quick_run` = `dont_schedule` ?

src/lib/run_environment.ml
```diff
@@ -94,10 +94,37 @@ module Tool = struct
   end
 end
 
+module Make_fun = struct
+  module Requirement = struct
+    type t = [
+      | `Processors of int
+      | `Internet_access
+      | `Memory of [ `GB of float | `Decent | `Big ]
+      | `Quick_run
```
ihodes: I think \`Quick_run means "shouldn't be scheduled"; why not have the converse as a requirement, e.g. \`Scheduler?

rleonid:
I can see how we are ignoring the requirements, but I'm still looking for an example of where and how we would enforce them.

ihodes:
@rleonid I think these would be enforced in e.g. Hamkepi's machine building modules like Demeter or Minerva, which know how our cluster looks and can determine how tools with various requirements should be run.

src/lib/run_environment.ml
rleonid: IMO, the benefits of a regular variant are: simplicity (no more what's the 'chick' for questions) and concise type definitions that use these variants.

rleonid:
Can you please provide an example?

src/lib/run_environment.ml
smondet: @rleonid  `dont_schedule` already makes weird assumptions (can the user actually "not" schedule? does schedule mean something their context?), the prevous API has `quick_run_program`, so `` `Quick_run`` replaced it.

@ihodes how would we put values for `` `GB _``? Right now I've been using the `` `Big`` and `` `Decent` because that's the only thing that has a tractable meaning.
Moreover using the combinators below make more sense `streamish` *is* “decent-memory requirement on one processor” (like `gunzip`, or `sam-to-bam`)

ihodes:
```ocaml
let run_program ?name ?processors ?requirements p =
    let open Ketrew.EDSL in
    let application_name =
      (* we sanitize the name at the last moment *)
      Option.value name ~default:"x"
      |> String.map ~f:(function
        | 'A' .. 'Z' | 'a' .. 'z' | '0' .. '9' as c -> c
        | other -> '_') in
   (* do a lot of matching on requirements to determine container_memory, processors, whether to run as a yarn_distributed_shell or just using the daemonize backend (matching on if it needs a cluster or internet access) etc. *)
    yarn_distributed_shell
      ~host
      ~container_memory
      ~timeout:(`Seconds (7 * 60 * 60 * 24))
      ~application_name
      ?container_vcores:processors
      Program.(p && sh "date" && sh "sleep 50" && sh "date")
      (* Ugly hack here; the sleep is to wait for NFS synchronization,
         if everything goes too fast, Ketrew may think that a target
         running on a compute node did not ensure the file because it
         checks for it on the login node. *)
```

smondet:
@rleonid something like

```ocaml
let run_program ?name ?(requirements=[]) prog =
    let open Ketrew.EDSL in
    begin match List.find_map requirements (function `Spark args -> Some args  | _ -> None) with
    | Some spark_args ->
        yarn_application (* .... *) prog
           ~jvm_crazy_args:["--gv7thGen=5GB"; "--zookeeperUrl=http://inside.the.network";....]
    | None ->
         if List.mem requirements `Quick_run || List.mem requirements `Internet
         then
             daemonize ~using:`Python_daemon ~host prog
         else
             yarn_distributed_shell prog
    end
```

src/lib/run_environment.ml
smondet: But then why would any job "not" be scheduled.

It's a `Machine.t`'s decision to say "I have a big, desert, login node, I can run some quick stuff here".

rleonid:
I think your example gives a facade of usefulness, but suffers from a weird inversion of control that I find frustrating. You will never write the total version of `run_program` that considers all possible subsets of requirements (assuming dedup-ing) because some combinations are confusing (ex.`` ~requirements:[`Spark; `Quick_run; `Processor 2 ]``, for idiots like me who want to calculate Pi with SparkPi). So you will have to consider them in some order, but here you are assuming that you know better how to satisfy those requirements than the user does.

I understand the sentiment of issue [#163](https://github.com/hammerlab/biokepi/issues/163), but I think the difference should be flushed out explicitly in different `make` types that do one thing concretely (run in shell, run on login node, run on spark) as opposed to munging all that logic into a list of requirements. This is all I will say on this issue.

smondet:
@rleonid just noting that ``[`Spark; `Quick_run; `Processor 2 ]`` has a valid meaning. As user, you can build a Ketrew backend plugin to run spark apps in non-yarn mode on some specific machine (its actually easy).
We need that inversion of control:
If you don't have that exact infrastructure; you can interpret the requirements *as best as you can* (e.g. run spark-on-yarn with 2 containers; and ignore the "quick-run" thing); or fail with a proper error message (`"Ugh, I am not a hadoop cluster; what is spark?"`).

> "you know better how to satisfy those requirements than the user does"

Who's `you`?

Here the Biokepi module is requiring stuff to run (XMas wish-list) and the `user` being the one who actually knows their infrastructure and creating a proper `Machine.t` that interprets the requirements.

ihodes:
LGTM—adding some docs like discussed offline, changing some names a bit, then merge away! Thanks for working this out.

ihodes:
LGTM! Thanks for adding docs :D

Issue #154, opened 2016-03-02
Set SM in bwa aligner


arahuja:
Currently in the BWA aligners (`aln` and `mem`) we set the `SM` or sample field as `SM`. 

https://github.com/hammerlab/biokepi/blob/master/src/lib/bwa.ml#L45-L47

Since every BAM file has `SM` set the same downstream tools don't properly set SM fields, i.e. Mutect will not give a column per sample in the VCF

iskandr:
Just hit this. 

Are we at least guaranteed that SM is always the normal sample? 

Also, it would be good to eventually get `TUMOR` and `NORMAL` as separate sample columns.

arahuja:
Are we at least guaranteed that SM is always the normal sample?
@iskandr funny enough, I don't think you are, it's random I think, as one just overwrites the other, but not sure.

PR #189, opened 2016-03-22
Pass sample ID to `bwa {aln,mem}`

smondet:
Cf. #154 

Do not merge, I haven't tested yet.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/189)

### DIFF
--- a/src/lib/bwa.ml
+++ b/src/lib/bwa.ml
@@ -34,7 +34,7 @@ let index
                && shf "bwa index %s"
                  (Filename.quote reference_fasta#product#path)))
 
-let read_group_header_option algorithm =
+let read_group_header_option algorithm sample_name =
   (* this option should magically make the sam file compatible
            mutect and other GATK-like pieces of software
            http://seqanswers.com/forums/showthread.php?t=17233
@@ -43,8 +43,10 @@ let read_group_header_option algorithm =
            `[bam_header_parse] missing LB tag in @RG lines.`
   *)
   match algorithm with
-  |`Mem -> "-R \"@RG\tID:bwa\tSM:SM\tLB:ga\tPL:Illumina\""
-  |`Aln -> "-r \"@RG\tID:bwa\tSM:SM\tLB:ga\tPL:Illumina\""
+  |`Mem ->
+    sprintf "-R \"@RG\tID:bwa\tSM:%s\tLB:ga\tPL:Illumina\"" sample_name
+  |`Aln ->
+    sprintf "-r \"@RG\tID:bwa\tSM:%s\tLB:ga\tPL:Illumina\"" sample_name
 
 let mem_align_to_sam
     ~reference_build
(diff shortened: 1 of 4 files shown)


arahuja:
Is this going to be a different value per fastq file then? Might need to do a `picard-tools AddOrReplaceReadGroups` at the end then?

smondet:
@arahuja (unless I missed something) If you pass the `~name` option you control which FASTQ files have the same sample-name; if you don't then yes they get each the name of their read-1.

smondet:
@arahuja ah yeah the problem is when we optimize we create new names:
https://github.com/hammerlab/biokepi/blob/67f37edb0561d26d39366f8715422684cd25b0ee/src/lib/pipeline.ml#L582-L583

arahuja:
@seb Yea, I see that, that makes sense. Just seems like it might get hidden that multiple fastqs aligned separately and then merged should update that field or make sure to set it properly originally. 

Some of kind constraint that says you can't (or at least I can't think of a use case) merge BAMs with different sample names

arahuja:
ah, ok, it also seem worthwhile to hang out to the information of where the read originally came from, so perhaps worth setting some of the other RG fields. Setting the ID to the fastq file name and setting the sample to field a simpler identifier?

arahuja:
It also seems that the setting of these fields will effect BQSR
> The recalibration system is read-group aware. It separates the covariate data by read group in the recalibration_report.grp file (using @RG tags)

smondet:
So right now it's `sprintf  "\"@RG\tID:bwa\tSM:%s\tLB:ga\tPL:Illumina\"" sample_name`

- `@RG` should get which value?
- `@ID` the individual FASTQ name?
- `@SM` the more global "sample-name" (≥ 1 FASTQ pairs)?

arahuja:
@seb yea that's what I was thinking. This page has some good information on how it will effect merging BAMs and BQSR:
http://gatkforums.broadinstitute.org/gatk/discussion/6472/read-groups

> Read group identifier
This tag identifies which read group each read belongs to, so each read group's ID must be unique. It is referenced both in the read group definition line in the file header (starting with @RG) and in the RG:Z tag for each read record.... In Illumina data, read group IDs are composed using the flowcell + lane name and number, making them a globally unique identifier across all sequencing data in the world.

> ID is the lowest denominator that differentiates factors contributing to technical batch effects: therefore, a read group is effectively treated as a separate run of the instrument in data processing steps such as base quality score recalibration, since they are assumed to share the same error model. There is another read group tag called PU (referring to the platform unit, holding flowcell lane information) that is not required by GATK but takes precedence over ID

> The name of the sample sequenced in this read group. GATK tools treat all read groups with the same SM value as containing sequencing data for the same sample, and this is also the name that will be used for the sample column in the VCF file. Therefore it's critical that the SM field be specified correctly.

Issue #34, opened 2015-05-02
Setup tool versioning


arahuja:
This is important for many of the tools as some versions are incompatible with different programs

A good example is `bwa` as we may want to specify a downstream tool requires a specific version.  A possible way to do this is to enumerate the versions supported by various tools and have the base tools setup the appropriate one.  I'll provide a code example shortly for what I'm thinking for feedback.

Not sure the best way to handle multiple versions of tools not installed by biokepi. If a workflow needs python 2 vs 3.

PR #194, opened 2016-03-23
Hla typing via seq2hla

rleonid:
This PR implements the RNA hla typing pipeline with seq2HLA. I recommend testing it with the `hla_typer.exe` command that now builds with `Biokepi`. The purpose of the command is not to be useful for actual HLA typing, but as testing that the full pipeline runs. I do not want to maintain this command... since we have an **e**DSL. The workflows, for `seq2HLA` depend on [biopam](https://github.com/solvuu/biopam) and [Conda](http://conda.anaconda.org/) and should install these tools appropriately.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/194)


src/lib/biopam.ml
```diff
@@ -4,114 +4,153 @@ Provide tools via Biopam: https://github.com/solvuu/biopam
...
-
-let seq2HLA_tool ~host ~meta_playground =
-  Tool.create Tool.Definition.(biopam "seq2HLA")
-    ~ensure:(install_seq2HLA ~host ~meta_playground)
-
-let toolkit ~host ~meta_playground () =
   Tool.Kit.create
-    [ picard_tool ~host ~meta_playground
-    ; seq2HLA_tool ~host ~meta_playground
+    [ mk (Library "PICARD_JAR") ~package:"picard"  ~witness:"picard.jar"
```
smondet: (not in this PR, just noting to that github cross-links stuff)
we should keep in mind that there are now 2 ways to get Picard in Biokepi (with potentially different versions)
cf. https://github.com/hammerlab/biokepi/issues/34

src/lib/pipeline.ml
```diff
@@ -84,9 +87,14 @@ type _ t =
...
   | Somatic_variant_caller: somatic Variant_caller.t * bam_pair t -> vcf t
   | Germline_variant_caller: germline Variant_caller.t * bam t -> vcf t
+  | Rna_hla_typer: fastq_sample t -> hla_types t
 
 module Construct = struct
 
+  (* Consider:
+     1. The name of this type makes more sense as a noun, fastq_input.
+     2. `Paired_end should contain (File.t * File.t) list ... since they are paired.
+     3. There needs to be a simpler type of `Single_pair and `Paired_type renamed. *)
```
smondet: This should be in an issue, not in comments.
(and “paired-end”/"single-end” seems to be the default nomenclature in bfx)

src/lib/pipeline.ml
```diff
@@ -84,9 +87,14 @@ type _ t =
   | Bam_pair: bam t * bam t -> bam_pair t
   | Somatic_variant_caller: somatic Variant_caller.t * bam_pair t -> vcf t
   | Germline_variant_caller: germline Variant_caller.t * bam t -> vcf t
+  | Rna_hla_typer: fastq_sample t -> hla_types t
```
smondet: Since this is Seq2HLA it should be in the actual name of the tag, idem for the output type and the constructor function.

src/lib/pipeline.ml
```diff
@@ -777,4 +792,21 @@ module Compiler = struct
     in
     compiler.wrap_gtf_node pipeline gtf_node
 
-end
+  let hla_types_step ~compiler (pipeline : hla_types pipeline) =
```
ihodes: Other compiler steps begin with `compile`; would be nice to be consistent.

src/lib/pipeline.ml
```diff
@@ -777,4 +792,21 @@ module Compiler = struct
...
+      | Paired_end_sample (dataset, l1, l2) ->
+          (* TODO: Seq2HLA can actually take the gzipped version too, so we'd
+             need a unique type for that. *)
+          let r1 = fastq_step ~read:(`R1 dataset) ~compiler l1 in
+          let r2 = fastq_step ~read:(`R2 dataset) ~compiler l2 in
+          let work_dir = work_dir // ("seq2HLA_wd_" ^ dataset) in
+          Seq2HLA.hla_type ~host:Machine.(as_host machine) ~work_dir
+            ~run_with:machine ~run_name:dataset ~r1 ~r2
+      | _ -> failwithf
+              "Seq2HLA doesn't support single_end_samples or paired end \
```
ihodes: -> "Seq2HLA doesn't support Single_end_samples or \
                Paired_end_samples of length greater than 1."

src/lib/pipeline.ml
```diff
@@ -777,4 +792,21 @@ module Compiler = struct
...
+                samples of length greater than 1."
```
smondet: right now `Paired_end_sample` will give you concatenated FASTQs so the length is already 1.

src/lib/pipeline.ml
```diff
@@ -392,6 +402,8 @@ let rec to_file_prefix:
       sprintf "%s-%s-%s" prev
         vc.Variant_caller.name
         vc.Variant_caller.configuration_name
+    | Rna_hla_typer s ->
+      sprintf "RNA_HLA_typer-%s" (to_file_prefix ?read s)
```
ihodes: This should be specific to the name of the typer (in this case, Seq2HLA); otherwise navigating nodes can't tell which typer is being run

src/lib/pipeline.ml
```diff
@@ -478,6 +490,8 @@ let rec to_json: type a. a t -> json =
           "Configuration", svc.Variant_caller.configuration_json;
           "Input", to_json bam_pair;
         ]]
+    | Rna_hla_typer input ->
```
ihodes: Similarly here, needs to have the name of the typer here

src/lib/seq2HLA.ml
```diff
@@ -0,0 +1,28 @@
+
+open Common
+open Run_environment
+module K = KEDSL
+
+let hla_type ?host ~work_dir ~run_with ~r1 ~r2 ~run_name =
```
ihodes: We can add this in another PR, but might be nice to expose configuration (if Seq2HLA has any) to the user.

src/lib/seq2HLA.ml
```diff
@@ -0,0 +1,28 @@
...
+  let tool = Machine.get_tool run_with (`Biopamed "seq2HLA") in
+  (* Why quote this here? Seems like it easy to create a bug,
+     why not enforce this at node construction ?*)
+  let r1pt = Filename.quote r1#product#path in
+  let r2pt = Filename.quote r2#product#path in
```
smondet: No need to change since this function is very short, but we prefer `Filename.quote` to be "close to" the `sh`/`shf`to avoid double escaping (happened before).

src/lib/workflow_utilities.ml
```diff
@@ -34,11 +34,11 @@ module Remove = struct
   (* This one is dirtier, it does not check its result and uses the `Host.t`
      directly, it should be used only when the `Machine.t` is not available
      (i.e. while defining a `Machine.t`). *)
-  let path_on_host ~host path =
+  let path_on_host ?host path =
```
smondet: Why this change? 
In Biokepi, the `~host` argument should be always there
(cf. https://github.com/hammerlab/biokepi/pull/170)

src/lib/pipeline.ml
rleonid: Addressed

src/lib/pipeline.ml
rleonid: I'm not so certain. There `Somatic_variant_caller` obviously does not refer to a specific tool. There seems to be a distinction between aligners and variant callers, where we use names in the first and what they do in the second. I think what a node of a pipeline does makes more sense, than the tool. What tool performs the specific step could be specified differently. Since `seq2HLA` is (at the moment) the only RNA typing tool. It is used by default.

src/lib/pipeline.ml
rleonid: These methods are also all in the `Compiler` module; `Compiler.compile_hla_types_step` seems redundant. Other methods (`fastq_step` and `fastq_sample_step`) were also factored out into separate functions.

src/lib/pipeline.ml
smondet: `Somatic_variant_caller` is a "factorization" of all the variant-callers 9just to make the GADT a bit less crazy), the constructors in `Construct` still use the name of the variants callers and `somatic Variant_caller.t` contains the SVC-specific info (including the "name" that will appear in the JSON).

We could do the same factorization for aligners indeed.

src/lib/pipeline.ml
rleonid: Addressed.

src/lib/pipeline.ml
rleonid: Fair enough, this was a carrier over of when I tried to match `Paired_end_sample (dataset, [l1], [l2])` when trying to rewrite those types, but I gave up on that. I'll remove the part about `Paired_end_sample`.

src/lib/pipeline.ml
rleonid: Please see message above about using `Rna_hla_typer` as the variant name. At the moment, there is only _one_ typer, so the message should suffice.

src/lib/seq2HLA.ml
rleonid: The "configuration" consists of hardcoded allele files is handled by biopam; there isn't any. Feel free to factor that out from our own [seq2HLA fork](https://bitbucket.org/hammerlab/seq2hla) and then just update the seq2HLA opam package [info](https://github.com/solvuu/biopam/tree/master/packages/seq2HLA/seq2HLA.2.2)

src/lib/seq2HLA.ml
rleonid: That is a 'preference'. But it would be _nicer_ if `ketrew` handled all these paths for you. I give you 50/50 odds that there is a filename quoting bug in Biokepi.

hammer: Can we add tests to cover quoting corner cases?

src/lib/pipeline.ml
smondet: We want the JSON and `Pipeline.Construct` (= what's user-level API) to be reasonably future-proof. We store those JSONs for later inspection.

The tag in the GADT decision should depend on whether we will have another HLA typer will the same I/O types as Seq2HLA; and as far as I understood, HLA typers don't have a standard output format.

src/lib/workflow_utilities.ml
rleonid: You're right, addressed.

src/lib/seq2HLA.ml
smondet: Ketrew has the `Program.exec` for that (escapes everything nicely).

Sometimes you cannot use `exec` because of some shell hack that you need for some tool (like `>` or `$VARIABLES`).

src/lib/pipeline.ml
rleonid: I'll address this issue in a future PR when there are actually different 'hla_typers'

src/lib/seq2HLA.ml
smondet: @hammer more tests would be awesome indeed `#HelpWanted`

(testing installation + run in normal use cases already takes lots of hours to people; I understand that trying corner cases and reading GATK/samtools corresponding error messages is often left out... espcially since some changes in, say, `bwa` can make semi-broken bams for BQSR much later in user pipelines)

src/lib/pipeline.ml
rleonid: To the extend that something needs to be "future-proof", there will exist a commit that will be aligned with the code that it is generating this JSON. So a "deserialization", to the extend that you may want one (even though, AFAIK, one currently doesn't exist), can be constructed. But since all you want is just more verbose and precise names and json, sure.

Addressed.

src/lib/pipeline.ml
```diff
@@ -324,9 +328,11 @@ module Construct = struct
        make_target }
       bam_pair
 
-end
+  let rna_hla_typer fastq_sample = Rna_hla_typer fastq_sample
```
smondet: This should also be `seq2HLA` for users of `Pipepline.Construct` to have future-proof and readable pipelines.

rleonid: I disagree, I think that the functions in `Pipeline.Construct` should be convenience wrappers for the elements of `Pipeline.t` GADT that they construct.

This distinction is similar, but different, from the aligner vs caller difference in the pipeline. At the moment, when you look at the variant caller "constructors"  in this module (ex `mutect`,  `muse`, `virmid`, `strelka` ... etc) they all have a similar type ( `: ~configuration:_ -> bam_pair t -> vcf t`) and a similar structure:

```OCaml
  let mutect ?(configuration=Mutect.Configuration.default) bam_pair =
    let configuration_name = configuration.Mutect.Configuration.name in
    let configuration_json = Mutect.Configuration.to_json configuration in
    let make_target
        ~reference_build ~run_with ~input ~result_prefix ~processors
        ?more_edges () =
      match input with | Variant_caller.Somatic {normal; tumor} ->
      Mutect.run
        ~configuration
        ?more_edges
        ~reference_build ~run_with
        ~normal ~tumor
        ~result_prefix `Map_reduce in
    somatic_variant_caller
      {Variant_caller.name = "Mutect";
       configuration_json;
       configuration_name;
       make_target;}
      bam_pair
``` 
vs
```OCaml
let mutect2 ?(configuration=Gatk.Configuration.Mutect2.default) bam_pair =
    let configuration_name = configuration.Gatk.Configuration.Mutect2.name in
    let configuration_json = Gatk.Configuration.Mutect2.to_json configuration in
    let make_target
        ~reference_build ~run_with ~input ~result_prefix ~processors
        ?more_edges () =
      match input with
      | Variant_caller.Somatic {normal; tumor} -> Gatk.mutect2
        ~configuration
        ?more_edges
        ~reference_build ~run_with
        ~input_normal_bam:normal ~input_tumor_bam:tumor
        ~result_prefix `Map_reduce in
    somatic_variant_caller
      {Variant_caller.name = "Mutect";
       configuration_json;
       configuration_name;
       make_target;}
      bam_pair
```
I think that this is bad because
1. it is biolerplate
2. why is it in `pipeline.ml` as opposed to `mutect.ml`? 

I think this limitation is because the `Pipeline.t` isn't further parameterized by the configuration of the underlying tools that it needs. Doing this would allow you to just use the `somatic_variant_caller` constructor, and the `Compiler` step will pattern match to the appropriate variant caller based on configuration.

I think you can make an argument that you want nicer, easier to read, construction functions (ie. `strelka`, `mutect` ..etc), but I would put those into a separate module (ie. `Pipeline.Specific` ?). Once another hla typer is added, I would make this distinction.

rleonid: To be specific, this new module (`Pipeline.Specific`) should probably be very simple redirects:

```OCaml
let mutect = Mutect.pipeline
let muse = Muse.pipeline
...
```

smondet: The code is in `Pipeline` because of the non-extensibility we discussed yesterday (Vs circular dependencies).

What's important is how this kind of **user** code looks like:
https://github.com/hammerlab/biokepi/blob/master/src/lib/common_pipelines.ml#L54-L58

If there was an `.mli` file the GADT would be hidden (or at least `private`); and the `Pipeline.Construct` module is the one that should really be exposed as “API”; we want the user to use functions that say which tool/configuration is used: `mutect`, `mutect2`, `somaticsniper` and `seq2HLA` because the tools even when they have the same goals; they do extremely different things and have very different results.

If a user starts using `Construct.rna_hla_typer` it means nothing, and looks like the "default" HLA typer may change in the future. Biokepi authors should not choose which tool is best/default for a task, that's up to the user (who most likely knows more biology than us, and what they actually want).

rleonid: > The code is in Pipeline because of the non-extensibility we discussed yesterday (Vs circular dependencies).

I don't think there is a necessary circular dependence. We can solve this problem the way that it is _always_ solved in OCaml. We can split `Pipeline` into a part where we declare the types, use those types in tool specific modules and then recombine them into a module containing shallow constructors ala `Pipeline.Specific` above.

> If a user starts using Construct.rna_hla_typer it means nothing, and looks like the "default" HLA typer may change in the future. Biokepi authors should not choose which tool is best/default for a task, that's up to the user (who most likely knows more biology than us, and what they actually want).

Agreed, but again, at this moment there is no support for another hla_typer (just like there is no support for another `gzip`). Why solve this problem until it is necessary. Feel free to file an issue if you think that we would forget.

smondet: `gunzip` is not a "kind" of tool, it an actual unix executable associated with a well defined format. If we want to start using `bzip2` we will call it `bzip2`.
 

My concern is that, as a user, I don't want to read `Construct.rna_hla_type` in any pipeline; and I don't want my pipelines to to fail (or do something different) in the future when Biokepi authors remove the function or, worse, change its meaning.

PR #196, opened 2016-03-24
Make `Picard.mark_duplicates` a “big” program

smondet:
On pretty standard data, mark-dups went way over 5 GB which is our current
heuristic.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/196)


ihodes:
LGTM

Issue #204, opened 2016-03-28
Strelka workflow is calling CombineVariants incorrectly


ihodes:
This gives us malformed VCFs. Of course.

PR #206, opened 2016-03-28
Merge Indels and SNVs properly for Strelka

ihodes:
Now we get just the two samples that we expect, TUMOR and NORMAL, in the header.

fixes https://github.com/hammerlab/biokepi/issues/204

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/206)

### DIFF
--- a/src/lib/strelka.ml
+++ b/src/lib/strelka.ml
@@ -163,11 +163,11 @@ let run ~reference_build
         && shf "cd %s" output_dir
         && shf "make -j%d" processors
         && Gatk.call_gatk ~analysis:"CombineVariants" [
-          "--variant"; "results/passed.somatic.snvs.vcf";
-          "--variant"; "results/passed.somatic.indels.vcf";
+          "--variant:snvs"; "results/passed.somatic.snvs.vcf";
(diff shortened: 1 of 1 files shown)


src/lib/strelka.ml
```diff
@@ -163,11 +163,11 @@ let run ~reference_build
...
         && Gatk.call_gatk ~analysis:"CombineVariants" [
-          "--variant"; "results/passed.somatic.snvs.vcf";
-          "--variant"; "results/passed.somatic.indels.vcf";
+          "--variant:snvs"; "results/passed.somatic.snvs.vcf";
+          "--variant:indels"; "results/passed.somatic.indels.vcf";
           "-R"; reference_fasta#product#path;
-          "-genotypeMergeOptions"; "UNIQUIFY";
-          "-o"; output_file_path;
+          "-genotypeMergeOptions"; "PRIORITIZE";
+          "-o"; output_file_path; "-priority"; "snvs,indels"
```
smondet: We should document somewhere what "`PRIORITIZE`" actually does, to make sure we don't loose indels on the way (i.e. what happens when there is an indel at the same location than an snv?).

PR #207, opened 2016-03-29
Expose a toolkit_install_path to run seq2HLA in conda environment

rleonid:
This PR enables the running of seq2HLA entirely inside of the biokepi conda environment. The problem that we experienced on demeter was due to `bowtie` itself being a python module (called via an env shebang) for argument processing. The problem was that if the default python wasn't compatible it would fail, while being called inside of `seq2HLA`. This presents two solutions:

1. Wrapping `bowtie` as well via `pyinstaller`, isolating its environment ... etc.
2. Giving up on that dream and exposing a working environment.

This PR takes the later approach. Certainly, the speed with which I am able to iterate on these tools is too slow for us to continue doing 1 for every tool. So an new approach is necessary.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/207)

### DIFF
--- a/src/app/hla_typer.ml
+++ b/src/app/hla_typer.ml
@@ -17,8 +17,10 @@ let pipeline_to_json ppln =
 let pipeline_to_workflow ~work_dir ?(uri="/tmp/ht") ppln =
   let open Pipeline.Compiler in
   let machine =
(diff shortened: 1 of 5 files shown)


smondet:
Discussed off-line.
Closing and fixing there: https://github.com/hammerlab/biokepi/pull/208

PR #208, opened 2016-03-29
Make `Conda` environment only “activable”

smondet:
The function `wrap_environment` was working only because Conda's deactivation
doesn't really work (at least it seems so).

This is kind-of a different solution from #207.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/208)


smondet:
I'm still testing by the way.

smondet:
It still works, but not by accident.

Issue #24, opened 2015-04-15
Make the reference-build a property of the BAM


arahuja:
Ran into this a few times but don't have a solution yet.  Most of the variant callers take a reference build and BAM, but those are coupled and it should not be possible to specify the wrong reference build for a BAM

arahuja:
Closed by #214

PR #214, opened 2016-03-30
Make `reference_build` part of `bam_file`

smondet:
This implements #24.

The function `KEDSL.transform_bam` also makes many workflows more readable and
safer.

We can later add a function `check_bams` that verifies at least one property: all the bams involved have the same ref-build (e.g. `normal` and `tumor` for variant callers; or the lists for bam-merging or indel-realigning).

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/214)

### DIFF
--- a/src/lib/common.ml
+++ b/src/lib/common.ml
@@ -94,18 +94,33 @@ module KEDSL = struct
 
   type bam_file = <
     is_done: Ketrew_pure.Target.Condition.t option;
+    host: Host.t;
     path : string;
     sorting: [ `Coordinate | `Read_name ] option;
-    content_type: [ `DNA | `RNA ];
+    reference_build: string;
   >
-  let bam_file ?host ?sorting ?(contains=`DNA) path : bam_file =
+  let bam_file ~host ?sorting ~reference_build path : bam_file =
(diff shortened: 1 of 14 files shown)


ihodes:
(LGTM, nice improvement)

arahuja:
This looks great, big fan of this! Is there any way in somatic calling to ensure the normal and tumor references are the same - that seems like a good failure mode to catch?

iskandr:
I like this idea but I'm surprised by the loose typing. How come
`reference_build` is a string rather than something more structured? Where
can I include information about patches (e.g. differentiate GRCh38.p1 from
GRCh38.p5)?

smondet:
@arahuja yes for a later PR, it's what I called `check_bams` above

@iskandr I think last time we discussed that topic, we just decided that for now different patches are completely different genomes (since it's easy to add/customize genomes now).
I'd be up for trying to put the reference-genome in a type-parameter, but in a later PR/experiment → https://github.com/hammerlab/biokepi/issues/215
(the change right-now is more contained and doesn't break any code I/we already have).

PR #216, opened 2016-03-31
Improve the build of the documentation

smondet:
This moves the build to the script `tools/build-doc.sh` to do the heavy duty.
It adds a webpage that includes 3 SVG dependency graphs from OCamlDoc's `-dot`
output.

screenshot 2016-03-31 13 22 36

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/216)


ihodes:
Nice! LGTM

PR #219, opened 2016-04-01
Adding Hisat2

arahuja:
This has been tested insofar in that it correctly produces the index and proper hisat2 command (alongside with hisat1) but the hisat2 alignment has not completed. I figured it was worth reviewing in the meantime

This follows the model we discussed previously:
- Creates a versioned Hisat in `run_environment.ml`
- Add a version argument to `hisat` in `tool_providers.ml` as well as putting the version in the install path
- Adds config to hisat to switch between versions

Let me know any critiques or if anything can be improved.

cc @smondet @ihodes

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/219)

### DIFF
--- a/src/lib/pipeline.ml
+++ b/src/lib/pipeline.ml
@@ -158,7 +158,7 @@ module Construct = struct
 
   let star fastq = Star fastq
 
-  let hisat ?(configuration = Hisat.Configuration.default) fastq = Hisat(configuration, fastq)
+  let hisat ?(configuration = Hisat.Configuration.default_v1) fastq = Hisat(configuration, fastq)
 
   let stringtie ?(configuration = Stringtie.Configuration.default) bam =
     Stringtie (configuration, bam)
--- a/src/lib/pipeline.ml
+++ b/src/lib/pipeline.ml
@@ -85,7 +85,7 @@ type _ t =
   | Gunzip_concat: fastq_gz t list -> fastq t
   | Concat_text: fastq t list -> fastq t
   | Star: fastq_sample t -> bam t
-  | Hisat: fastq_sample t -> bam t
+  | Hisat: Hisat.Configuration.t * fastq_sample t -> bam t
   | Stringtie: Stringtie.Configuration.t * bam t -> gtf t
   | Bwa: bwa_params * fastq_sample t -> bam t
   | Bwa_mem: bwa_params * fastq_sample t -> bam t
@@ -158,7 +158,7 @@ module Construct = struct
 
   let star fastq = Star fastq
 
-  let hisat fastq = Hisat fastq
+  let hisat ?(configuration = Hisat.Configuration.default) fastq = Hisat(configuration, fastq)
 
   let stringtie ?(configuration = Stringtie.Configuration.default) bam =
     Stringtie (configuration, bam)
@@ -371,10 +371,10 @@ let rec to_file_prefix:
         (to_file_prefix sample) gap_open_penalty gap_extension_penalty
     | Star (sample) ->
       sprintf "%s-star-aligned" (to_file_prefix sample)
-    | Hisat (sample) ->
-      sprintf "%s-hisat-aligned" (to_file_prefix sample)
+    | Hisat (conf, sample) ->
+      sprintf "%s-hisat-%s-aligned" (to_file_prefix sample) (conf.Hisat.Configuration.name)
     | Stringtie (conf, sample) ->
-      sprintf "%s-%sstringtie"
+      sprintf "%s-%s-stringtie"
         (to_file_prefix sample)
         (conf.Stringtie.Configuration.name)
     | Mosaik (sample) ->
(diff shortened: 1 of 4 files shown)


src/lib/run_environment.ml
```diff
@@ -54,7 +55,7 @@ module Tool = struct
     let star = custom "star" ~version:"2.4.1d"
     let stringtie = custom "stringtie" ~version:"1.2.2"
     let cufflinks = custom "cufflinks" ~version:"2.2.1"
-    let hisat = custom "hisat" ~version:"0.1.6-beta"
+    let hisat = `Hisat `V_2_0_2_beta
```
ihodes: I think this is good; latest tools should be the default.

src/lib/tool_providers.ml
```diff
@@ -494,7 +506,7 @@ let default_toolkit
     star_tool ~host ~meta_playground;
     stringtie_tool ~host ~meta_playground;
     cufflinks_tools ~host ~meta_playground;
-    hisat_tool ~host ~meta_playground;
+    hisat_tool ~host ~meta_playground ~version:`V_0_1_6_beta;
```
ihodes: This is my confusion around tools again; why V2 in the run_environment but V1 in the default toolkit?

arahuja: Yea, actually was just thinking that - the two are not really linked. Here I create `V_0_1_6_beta` tool, and there I say if you ask for the default it's the V2 one (which may not be in the kit unless you add it)

ihodes: Maybe have both tools in the toolkit, but the default tool should remain V2?

arahuja: I guess the question is should there be a 1-1 mapping between the default tools and the default toolkit - if so, how is it enforced

arahuja: > Maybe have both versions in the toolkit, but the default tool should remain V2?

I thought have 2 versions of a single tool in the default kit seemed weird, I added V2 to the kit when I used it.

ihodes: I think it's fine to have other versions in the default toolkit. What seems more confusing is saying that the default tool does not exist in the default toolkit.

ihodes: All the nodes in the default toolkit are doing is telling the machine how to get the tool, right?

src/lib/tool_providers.ml
```diff
@@ -111,12 +111,24 @@ let star_tool ~host ~meta_playground =
...
   let archive = Filename.basename url in
-  let hisat_binary = "hisat" in
+  let hisat_binary = 
+    match version with
+    | `V_0_1_6_beta -> "hisat"
+    | `V_2_0_2_beta -> "hisat2"
+  in
   let ensure =
     workflow_node (single_file ~host (install_path // hisat_binary))
       ~name:(sprintf "Install HISAT")
```
ihodes: Maybe we should add the version to the name?

src/lib/pipeline.ml
```diff
@@ -158,7 +158,7 @@ module Construct = struct
 
   let star fastq = Star fastq
 
-  let hisat fastq = Hisat fastq
+  let hisat ?(configuration = Hisat.Configuration.default_v1) fastq = Hisat(configuration, fastq)
```
smondet: Shouldn't `v2` be the default?

src/lib/tool_providers.ml
smondet: Yes, having both version in the default tool-kit is what we want. They will be installed only when used anyway.

src/lib/pipeline.ml
arahuja: If you'd like? My assumption was the default should be what was there already so nothing changes for any pipelines using v1

smondet: Actually maybe :)

I don't know how the two versions are different, if you think users expect to very different results then yes `v1` is better.

arahuja:
@smondet @ihodes not sure if you are still reviewing, but from the feedback so far, my takeaways are:

1. Add both to the Toolkit.default
2. Make the default tool in run environment and default config in pipeline match and that should be ... `V2`?

ihodes:
1. for sure (no downside)
2. i don't have a preference; seems good to keep things recent, but also have to watch out for that e.g. I need to switch the PGV pipeline code if we make this change.

arahuja:
> i don't have a preference; seems good to keep things recent, but also have to watch out for that e.g. I need to switch the PGV pipeline code if we make this change.
@ihodes For this case, it doesn't matter really as we likely know all end-users of this, but in general, I think major upgrades (especially when the previous version isn't broken) should be opt-in rather than opt-out.  In this case, if you using `hisat-v1` I think it's worthwhile to stay with until we have some time to sanity check and compare v2

smondet:
@arahuja yes to both !

PR #221, opened 2016-04-04
Add OptiType invocation.

rleonid:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/221)

PR #227, opened 2016-04-05
Use bioconda pysam version that comes with friends.

rleonid:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/227)


smondet:
Thanks, merged into #221.

PR #228, opened 2016-04-08
Split library in 5

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/228)


README.md
```diff
@@ -1,14 +1,27 @@
...
 
-The library also contains an experimental module called `Biokepi_pipeline` which
-uses a GADT to concisely express well-typed analysis pipelines. A *demo* command
-line application `biokepi` demonstrates how to run such a pipeline.
-The library can download most of the tools by itself (through more Ketrew
-targets).
+
+Biokepi is split into 4 libraries:
+
+- `Biokepi_run_environment`: is the common API used all accross the library
```
ihodes: accross -> across

ihodes:
LGTM! Thanks for splitting it up

src/lib/biokepi.ml
```diff
@@ -0,0 +1,15 @@
+
+module Pipeline = Biokepi_pipeline_edsl.Pipeline
+
+module KEDSL = Biokepi_run_environment.Common.KEDSL
+
+module Metadata = Biokepi_run_environment.Metadata
+
+module Machine = Biokepi_run_environment.Machine
+
+module Setup = struct
```
ihodes: why is this != `module Setup = Biokepi_environment_setup`?

smondet: I actually started with more than one thing in there, and then I switched back to one module, but I wanted to see what `ocamldoc` does differently in both cases, and then I forgot :)

I can change it if you think the homogeneity is worth it. That module is going to change soon anyways, as I will figure out a way to generate better documentation.

PR #230, opened 2016-04-12
Ask for more memory when running `samtools sort`

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/230)

### DIFF
--- a/src/bfx_tools/samtools.ml
+++ b/src/bfx_tools/samtools.ml
@@ -114,7 +114,9 @@ let tabix ~run_with ~tabular_format input_file =
 
 let do_on_bam
     ~(run_with:Machine.t)
-    ?(more_depends_on=[]) ~name input_bam ~product ~make_command =
+    ?(more_depends_on=[]) ~name
+    ?(more_requirements: Machine.Make_fun.Requirement.t list = [])
+    input_bam ~product ~make_command =
   let open KEDSL in
   let samtools = Machine.get_tool run_with Machine.Tool.Default.samtools in
   let src = input_bam#product#path in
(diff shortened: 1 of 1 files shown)


src/bfx_tools/samtools.ml
```diff
@@ -150,6 +154,7 @@ let sort_bam_no_check ~(run_with:Machine.t) ?(processors=1) ~by input_bam =
     | `Read_name -> "sort" :: "-n" :: command
   in
   do_on_bam ~run_with input_bam ~product ~make_command
+    ~more_requirements:[`Memory `Big]
```
arahuja: This must be new - what is ` \`Big`?  Seems pretty relative, no?

smondet: Yes it is. It is the temporary way of saying "a lot of mem". One day we will try to find out the exact requirements, but that requires a lot of work. 
(memory usage is not even always a monotonous function of input size, for example, maximal coverage in a region of a bam can make tools create big pileups, or big gaps between paired reads, etc.)

arahuja: Yea, my thinking was since it's hard to tell, making it an explicit option might easier. `Big` to this might be small or not even possible on other machines. Where is Big translated to a number?

ihodes: This is another reason to let users easily tweak their tools' memory requirements themselves :)

smondet: @arahuja `` `Big`` (or any requirement) is instanciated by the user when they setup their `Machine.t`.

PR #234, opened 2016-04-13
Add support for “self-identification”

### LINKED ISSUE #213: Add ID specific to each tool/command which Machine.t can dispatch on
Sometimes we (pipeline writers) know better than Biokepi writers.

smondet:
This fixes #213.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/234)

### DIFF
--- a/src/bfx_tools/bwa.ml
+++ b/src/bfx_tools/bwa.ml
@@ -30,6 +30,7 @@ let index
     ]
     ~tags:[Target_tags.aligner]
     ~make:(Machine.run_big_program run_with ~processors:1 ~name
+             ~self_ids:["bwa"; "index"]
              Program.(
                Machine.Tool.(init bwa_tool)
                && shf "bwa index %s"
@@ -98,6 +99,7 @@ let mem_align_to_sam
         :: [])
       ~tags:[Target_tags.aligner]
       ~make:(Machine.run_big_program run_with ~processors ~name
+               ~self_ids:["bwa"; "mem"]
                Program.(
                  Machine.Tool.(init bwa_tool)
                  && in_work_dir
@@ -166,6 +168,7 @@ let align_to_sam
       ]
       ~tags:[Target_tags.aligner]
       ~make:(Machine.run_big_program run_with ~processors ~name
+               ~self_ids:["bwa"; "aln"]
                Program.(
                  Machine.Tool.(init bwa_tool)
                  && in_work_dir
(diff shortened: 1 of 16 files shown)


src/run_environment/machine.ml
```diff
@@ -122,6 +122,11 @@ module Make_fun = struct
                                   ["#WIP"]). *)
       | `Custom of string (** Pass arbitrary data (useful for temporary
                               extensions/experiements outside of Biokepi). *)
+      | `Self_identification of string list
```
ihodes: This might be nicer to match with downstream as just `Self_identification of string` instead. 

We could also use a type instead. If we used a type instead of a string, Merlin/the compiler could help us figure out what we can match on. As it is, someone will have to go through the source code to figure out what they can match on (and it could change/break on them in the future without a warning).

(Still not sure, even with the Hamkepi PR, that this solution is as simple for the user as passing `?requirements` down from the tool itself…)

src/bfx_tools/bwa.ml
```diff
@@ -219,6 +222,7 @@ let align_to_sam
       (single_file result ~host:Machine.(as_host run_with))
       ~name ~edges
       ~tags:[Target_tags.aligner]
-      ~make:(Machine.run_big_program run_with ~processors:1 ~name  program)
+      ~make:(Machine.run_big_program run_with ~processors:1 ~name  program
+               ~self_ids:["bwa"; "sampe"])
```
ihodes: What is `sampe`?

smondet: It's the name of the `bwa` command that makes a `.sam` out of alignments.

src/run_environment/machine.ml
smondet: It would be very annoying to manage:

- the `?requirements` (the argument needs to go up the stack very far, you would end-up with `Gatk.indel_realigner` that takes tons of overriding arguments: `?gatk_requirements`, `?samtools_sort_requirements`, `?samtools_index_requirements`, `?merge_bams_requirements`, etc. (installations, downloads, …)
- and then it would have to go up the `Pipeline_edsl` API too

And on top of that it would break the clean-ish separation between bioinformatics and infrastructure. The goal is to be able to reuse 100% of the pipeline/`Tools` code unchanged, and switch between completely different `Machine.t`s at will.

ihodes: Another downside: if you have different invocations of the same tool, you can't configure them separately.

ihodes:
Tough to name this, but instead of `~self_ids`, we could just have `~ids` (as it is the ID of the make process), and the type in `~requirements` could be `` `Make_process_id``? 

(I can file this as a separate issue)

PR #235, opened 2016-04-14
Add @RG to star

ihodes:
Complies with SAM spec (must have a RG specified) and ensures this works
with the downstream GATK.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/235)

### DIFF
--- a/src/bfx_tools/star.ml
+++ b/src/bfx_tools/star.ml
@@ -77,7 +77,9 @@ let align
       (Filename.quote star_index_dir)
       processors
       result_prefix
-      (sprintf "ID:%s" (Filename.basename r1_path))
+      (sprintf "ID:%s SM:\"%s\""
+         (Filename.basename r1_path)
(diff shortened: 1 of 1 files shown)


ihodes:
Don't merge, need to add the SM field, at least, so GATK doesn't NPE on the output of this BAM: 

http://gatkforums.broadinstitute.org/gatk/discussion/7282/gatk-depthofcoverage-runtimeerror-nullpointerexception

smondet:
Maybe not for this PR, but worth knowing that we should aggregate (and then use ☺) this information in the `fastq_reads` product:
https://github.com/hammerlab/biokepi/blob/master/src/run_environment/common.ml#L72
(here we use the `sample_name` but not the `fragment_id` yet:
https://github.com/hammerlab/biokepi/blob/master/src/bfx_tools/bwa.ml#L106).

PR #236, opened 2016-04-14
TTFI-pipeline: 2nd strike

smondet:
New things:

- Compilers `To_json` and `To_dot`
- Generic `Optimization_framework` (from QueL)
- One optimization pass that has mostly “display-value”

Not sure if `Biokepi.EDS.Transform.Beta_reduce` can be called like that. This is what the test outputs:

Pipeline:

pipeline-1 dot

Pipeline with `Transform_applications` applied:

pipeline-1-beta

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/236)

### DIFF
--- a/src/lib/biokepi.ml
+++ b/src/lib/biokepi.ml
@@ -14,6 +14,13 @@ module EDSL = struct
        and
        type 'a observation = Yojson.Basic.json =
       Biokepi_pipeline_edsl.To_json
+
+    module To_dot : Semantics.Bioinformatics_base
+      with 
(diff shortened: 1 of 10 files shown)


smondet:
This starts addressing #84 (but the EDSL itself is incomplete for now).

hammer:
@smondet very cool! any chance we could get @pveber to do some code reviews to be certain he'll be able to make use of ttfi? Having an external user for ketrew/biokepi would be wonderful.

PR #239, opened 2016-04-17
Add Optitype to the pipeline

ihodes:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/239)

### DIFF
--- a/src/pipeline_edsl/pipeline.ml
+++ b/src/pipeline_edsl/pipeline.ml
@@ -889,7 +889,7 @@ module Compiler = struct
            need a unique type for that. *)
         let r1 = fastq_step ~read:(`R1 info.fragment_id) ~compiler l1 in
         let r2 = fastq_step ~read:(`R2 info.fragment_id) ~compiler l2 in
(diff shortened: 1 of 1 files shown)


src/pipeline_edsl/pipeline.ml
```diff
@@ -885,4 +898,21 @@ module Compiler = struct
...
 
+  let rec optitype_hla_types_step ~compiler (pipeline : optitype_hla_types pipeline) =
+    let { machine ; work_dir; _ } = compiler in
+    match pipeline with
+    | Optitype (kind, sample) ->
+      begin match sample with
+      | Paired_end_sample (info, l1, l2) ->
+        let r1 = fastq_step ~read:(`R1 info.fragment_id) ~compiler l1 in
+        let r2 = fastq_step ~read:(`R2 info.fragment_id) ~compiler l2 in
+        let work_dir = work_dir // ("optitype_wd_" ^ info.fragment_id) in
```
smondet: instead of `optitype_wd_` we should use `to_result_prefix` so that running with `` `DNA`` or `` `RNA`` with the same FASTQs does not result in a clash.

ihodes:
Fixed!

PR #244, opened 2016-04-19
Add exome annotation to mm10 reference

ihodes:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/244)

### DIFF
--- a/src/run_environment/reference_genome.ml
+++ b/src/run_environment/reference_genome.ml
@@ -175,6 +175,9 @@ module Default = struct
     let dbsnp_mm10_indels_url =
       "ftp://ftp-mouse.sanger.ac.uk/REL-1303-SNPs_Indels-GRCm38/\
        mgp.v3.indels.rsIDdbSNPv137.vcf.gz" in
+    let gene_annotations_gtf =
+      "ftp://ftp.ensembl.org/pub/release-79/gtf/mus_musculus/\
+       Mus_musculus.GRCm38.79.gtf.gz" in
     create Name.mm10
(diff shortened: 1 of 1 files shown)


src/run_environment/reference_genome.ml
```diff
@@ -175,6 +175,9 @@ module Default = struct
     let dbsnp_mm10_indels_url =
       "ftp://ftp-mouse.sanger.ac.uk/REL-1303-SNPs_Indels-GRCm38/\
        mgp.v3.indels.rsIDdbSNPv137.vcf.gz" in
+    let gene_annotations_gtf =
+      "ftp://ftp.ensembl.org/pub/release-79/gtf/mus_musculus/\
```
iskandr: Out of curiosity, why release 79 and not something more recent?

ihodes: Pretty arbitrary; we're using 79 in other places. We need a better story when it comes to minor releases of the same genome...

ihodes: https://github.com/hammerlab/biokepi/issues/245

iskandr: Just a small note: an Ensembl release doesn't necessarily correspond to a
different minor patch of a genome. They release on a regular schedule and
sometimes it's just an update to the genome metadata (e.g. transcript
intervals, biotype of a gene, &c) while still using the exact same genome
sequence as the previous release.

ihodes: Noted, thank you

PR #248, opened 2016-04-20
Allow seq2HLA to run on Bam_to_fastq 

arahuja:
Updating the `seq2HLA` pipeline to run on fastqs from BAMs. Let me know if there was a better way to achieve this.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/248)

### DIFF
--- a/src/pipeline_edsl/pipeline.ml
+++ b/src/pipeline_edsl/pipeline.ml
@@ -927,7 +927,7 @@ module Compiler = struct
         let work_dir = work_dir // (to_file_prefix pipeline) ^ "_work_dir" in
         Optitype.hla_type
           ~work_dir ~run_with:machine ~run_name:info.fragment_id ~r1 ~r2 kind
-      | _ -> failwithf "Seq2HLA doesn't support Single_end_sample(s)."
+      | _ -> failwithf "Optitype doesn't support Single_end_sample(s)."
       end
     | With_metadata (metadata_spec, p) ->
       optitype_hla_types_step ~compiler p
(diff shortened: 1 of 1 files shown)


src/pipeline_edsl/pipeline.ml
```diff
@@ -899,6 +899,17 @@ module Compiler = struct
         let work_dir = work_dir // (to_file_prefix pipeline) ^ "_work_dir" in
         Seq2HLA.hla_type
           ~work_dir ~run_with:machine ~run_name:info.fragment_id ~r1 ~r2
+      | Bam_to_fastq(info_opt, how, what) -> 
+        let fastq_pair = fastq_sample_step ~compiler sample in
```
smondet: This line is the right way of compiling the whole `sample` part, there is no need to pattern-match on `sample` any more.

src/pipeline_edsl/pipeline.ml
```diff
@@ -916,7 +927,7 @@ module Compiler = struct
         let work_dir = work_dir // (to_file_prefix pipeline) ^ "_work_dir" in
         Optitype.hla_type
           ~work_dir ~run_with:machine ~run_name:info.fragment_id ~r1 ~r2 kind
```
smondet: And the same thing can be done here :)

src/pipeline_edsl/pipeline.ml
arahuja: Ah yea, I see what you mean. I had thought that it would not give me `fragment_id` as needed above, but I see that it does

arahuja:
@smondet Added commit https://github.com/hammerlab/biokepi/pull/248/commits/bd350649b51cef9a03a0a71aac61f3dea4beae3f is that what you meant?

smondet:
@arahuja yes, thanks! LGTM, will merge soon.

PR #250, opened 2016-04-21
Add path to optitype product

ihodes:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/250)


src/bfx_tools/optitype.ml
```diff
@@ -41,6 +41,7 @@ let hla_type ~work_dir ~run_with ~r1 ?r2 ~run_name nt =
     in
     object
       method is_done = Some (`Volume_exists vol)
+      method path = work_dir
```
smondet: You put the path to the directory because optitype creates new filenames everytime right?

ihodes: I don't actually know what it creates (haven't gotten it to run yet…)

PR #254, opened 2016-04-26
TTFI-EDSL: Pipeline Library + 1 example extension

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/254)

### DIFF
--- a/src/examples/edsl_extension_register_result.ml
+++ b/src/examples/edsl_extension_register_result.ml
@@ -347,20 +347,25 @@ let tumor =
     ]
   )
 (*M
+
 We submit the workflow to the Ketrew server (here using the `default`
 configuration):
 M*)
 let () =
-  run ~machine:My_cluster.machine
-    ~work_dir:My_cluster.work_dir
-    ~sqlite_db:(My_cluster.work_dir // "test-biokepi.sqlite") ()
-    ~processors:My_cluster.max_processors
-    ~normal ~tumor
-  |> Ketrew.Client.submit_workflow
+  match Sys.argv.(1) with
+  | "run" ->
+    run ~machine:My_cluster.machine
+      ~work_dir:My_cluster.work_dir
+      ~sqlite_db:(My_cluster.work_dir // "test-biokepi.sqlite") ()
+      ~processors:My_cluster.max_processors
+      ~normal ~tumor
+    |> Ketrew.Client.submit_workflow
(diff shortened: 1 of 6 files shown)


smondet:
Result of the example.

smondet:
Also, I pushed the HTML there: http://wr.mondet.org/smondet/biokepi-doc/

src/examples/edsl_extension_register_result.ml
```diff
@@ -0,0 +1,394 @@
...
+     end)
+= struct
+  include Biokepi.EDSL.Compile.To_workflow.Make(Config)
+  open Biokepi.EDSL.Compile.To_workflow.File_type_specification
+  let register : type a . string -> a t -> a t = fun metadata x ->
+    let make_node any_node =
+      let registration =
+        let path = any_node#product#path in
+        let ketrew_id = Ketrew.EDSL.node_id any_node in
+        let json = Mem.look_up metadata in
```
ihodes: Why/when is a JSON DS being saved?

smondet: There: https://github.com/hammerlab/biokepi/pull/254/files#diff-e9250c1eb1e5cc0ecbfcc6179bc542a3R151

ihodes: So the json is only available if you first run the `To_json_with_register` compiler?

smondet: Yes; which makes sense :)

ihodes: Might be good to have a note in the docs above this explaining that. How simple would it be for the register node to do the JSON compilation on the nodes it's interested in (so the preprocessing step isn't required)?

smondet: The things is that at this point we only require that `look_up` function (explained above). The fact that we provide `Mem` using the compiler `To_json` by calling is a decision taken later.

This is also valid for example:

```ocaml
module M = struct
 let look_up s = `Assoc ["tag", `String s]
end
```

I think I see how to make both step happening together, but it seems it would be adding complexity while the example script is already pretty long.

ihodes: Cool. I'd definitely want to be able to do both in the same compiler pass in the PGV workflow, for example.

ihodes:
LGTM!

PR #257, opened 2016-04-28
Fix optitype

ihodes:
Use fastq workflow_node in lieu of the r1/r2 components of one, and
correctly depend on it.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/257)


smondet:
Thanks 👍

PR #256, opened 2016-04-28
Optional configurations

smondet:
This PR is on top of #254, and depends on https://github.com/hammerlab/ketrew/pull/433.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/256)

PR #258, opened 2016-04-28
Make the optional arguments truly optional

smondet:
This was an oversight in #256; now the variant callers take one `unit`
argument.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/258)


ihodes:
LGTM

PR #259, opened 2016-04-28
Add overhang parameter to STAR

ihodes:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/259)

### DIFF
--- a/src/bfx_tools/star.ml
+++ b/src/bfx_tools/star.ml
@@ -10,13 +10,19 @@ module Configuration = struct
     type t = {
       name: string;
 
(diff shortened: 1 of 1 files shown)


arahuja:
You are not actually using it here?

ihodes:
Not by default; I just want the option to be able to use it (I can set it in PGV)

arahuja:
Ah I didn't notice the new 
```
       (Configuration.Align.render configuration
        |> String.concat ~sep:" ")		        |> String.concat ~sep:" ")
```
to parse out any new config

PR #274, opened 2016-05-23
Minor documentation fixes

smondet:
The build of the documentation is still kinda broken at least it does not fail now.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/274)


armish:
👏 thanks for adding that link!

PR #282, opened 2016-06-02
Fix some issues with Fastqc

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/282)

### DIFF
--- a/src/environment_setup/tool_providers.ml
+++ b/src/environment_setup/tool_providers.ml
@@ -6,13 +6,13 @@ let rm_path = Workflow_utilities.Remove.path_on_host
 
 let generic_installation
     ~(run_program : Machine.Make_fun.t)
-    ~host ~install_path 
+    ~host ~install_path
     ~install_program ~witness ~url
     ?unarchived_directory
     tool_name =
   let archive = Filename.basename url in
-  let archive_kind = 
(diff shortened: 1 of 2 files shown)


ihodes:
LGTM

armish:
thanks, @smondet! LGTM, too.

Not sure how all those whitespaces crept into the code :\

src/environment_setup/tool_providers.ml
```diff
@@ -409,7 +411,8 @@ let fastqc =
         shf "cp -r * %s" path
         && shf "chmod +x %s" (binary_path path)
       )
-  ~init_program
+    ~init_program
+    ~unarchived_directory:"FastQC"
```
armish: btw, why do we need this?

smondet: The default value for that is `<name-of-the-tool>*`:
https://github.com/hammerlab/biokepi/blob/62e810311228decd51ab5bb0ef0f5e2638262828/src/environment_setup/tool_providers.ml#L26

And fastqc's zip file seems to unzip to a `FastQC/` directory.

So without the option it would work on MacOSX (“case aware” craziness) but not on a normal filesystem.

armish: gotcha 👍

PR #284, opened 2016-06-06
Add particular paths to seq2HLA results

ihodes:
We want to manipulate these files later in the pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/284)

### DIFF
--- a/src/bfx_tools/seq2HLA.ml
+++ b/src/bfx_tools/seq2HLA.ml
@@ -2,7 +2,16 @@
 open Biokepi_run_environment
 open Common
 
-let hla_type ~work_dir ~run_with ~r1 ~r2 ~run_name =
+
+type product = <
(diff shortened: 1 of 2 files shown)


src/bfx_tools/seq2HLA.ml
```diff
@@ -2,6 +2,14 @@
...
 
+
+type product = <
+    is_done : Ketrew_pure.Target.Condition.t option ;
+    class1_path : string;
+    class2_path: string;
+    work_dir_path: string >
+
+
 let hla_type ~work_dir ~run_with ~r1 ~r2 ~run_name =
```
smondet: LGTM but since we don't have an MLI file there; could you force the type with an annotation (`: product KEDSL.workflow_node`)?

ihodes: Good idea; just added that.

PR #286, opened 2016-06-07
Implement map-reduction of BQSR (fix #184)

### LINKED ISSUE #184: Implement parallelization for BQSR
Cf. http://gatkforums.broadinstitute.org/wdl/discussion/1988/a-primer-on-parallelism-with-the-gatk

and http://gatkforums.broadinstitute.org/gatk/discussion/1919/parallelizing-base-quality-score-recalibration for specifics wrt BQSR

smondet:
Still testing.

PS: sorry for the whitespacy diff 🙏

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/286)

### DIFF
--- a/src/bfx_tools/gatk.ml
+++ b/src/bfx_tools/gatk.ml
@@ -486,7 +486,7 @@ let base_quality_score_recalibrator_on_region
           "-knownSites"; db_snp#product#path;
           "-o"; recal_data_table;
         ] @ Configuration.Bqsr.render bqsr_configuration)
-        && call_gatk ~analysis:"PrintReads" ([
+        && call_gatk ~analysis:"PrintReads" ~region ([
           "-nct"; Int.to_string processors;
           "-R"; fasta#product#path;
           "-I"; sorted_bam#product#path;
--- a/src/bfx_tools/gatk.ml
+++ b/src/bfx_tools/gatk.ml
@@ -447,11 +447,87 @@ let call_gatk ~analysis ?(region=`Full) args =
   sh (String.concat ~sep:" "
         ("java -jar $GATK_JAR -T " :: analysis :: intervals_option :: escaped_args))
 
-let base_quality_score_recalibrator
+let base_quality_score_recalibrator_on_region
     ~configuration:(bqsr_configuration, print_reads_configuration)
     ~run_with
-    ~input_bam ~output_bam =
+    ~input_bam
+    ~result_prefix
+    ~region =
   let open KEDSL in
+  let region_name = Region.to_filename region in
+  let output_file suffix =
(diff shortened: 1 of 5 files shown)


src/bfx_tools/gatk.ml
```diff
@@ -497,6 +505,25 @@ let base_quality_score_recalibrator
...
+    Machine.get_reference_genome run_with input_bam#product#reference_build in
+  let run_on_region region =
+    base_quality_score_recalibrator_on_region ~configuration
+      ~run_with ~input_bam ~result_prefix ~region
+  in
+  match how with
+  | `Region region -> run_on_region region
+  | `Map_reduce ->
+    let all_nodes =
+      List.map (Reference_genome.major_contigs reference)
```
ihodes: Does this mean it'll not run on other regions? Seems important that we don't leave those behind (I'm guessing that we don't run other parallelized tools over non-major-contigs?)

smondet: Yes all region-based map-reducing uses the `major_contigs` function (then the user can put whatever they want behind that; it's just a property of the reference genome).

arahuja:
Doesn't this throw off  the calibration? Is there enough data within a chromosome for all of the covariates? And this will give different results compared to running all together?

Perhaps the first `BaseRecalibrator` step should be run on all the data and then `PrintReads` should be run in parallel?

ihodes:
@arahuja Broad seems to think it's okay: https://github.com/hammerlab/biokepi/issues/184

src/bfx_tools/gatk.ml
ihodes: Is there an easy way for the user to specify that it should be run over all the contigs? Or would they need to enumerate the name of all the other small contigs/unmapped regions in the reference in order to get this behavior? (And thereby creating a "new" `Reference_genome.t`?)

arahuja:
@ihodes commented at https://github.com/hammerlab/biokepi/issues/184#issuecomment-224674310

smondet:
@ihodes @arahuja Also, I don't know if related but with the PR my strelka fails with: https://sites.google.com/site/strelkasomaticvariantcaller/home/faq#TOC-The-workflow-reported-an-error-described-as-ERROR:-multiple-suggested-alignments-for-read-are-not-same-strand-

smondet:
@ihodes @arahuja so we confirm that this //ization method (= per contig) goes nowhere?

ihodes:
Yeah unfortunately I think we should avoid merging this, and punt until we can work out how to properly parallelize it with their tools.

smondet:
OK 😿

ihodes:
😢

PR #289, opened 2016-06-16
Fix destructuring of FASTQ reads

smondet:
The implementation of the `#r1` and `#r2` methods was cutting the
dependency chain. Now the “product-level” methods return products and
there are separate functions at the “node-level” which preserve
dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/289)


ihodes:
LGTM

PR #290, opened 2016-06-24
Allow link/copy of input files

### LINKED ISSUE #112: Input Bams should be linked (`ls -s`)
Here https://github.com/hammerlab/biokepi/blob/master/src/lib/pipeline.ml#L511

If we give bams that are not indexed, `samtools` will index them in place which breaks the assumption that Biokepi does everything inside its playground.

smondet:
This is in `EDSL.To_workflow`, a compiler option decides whether to
copy/link the input files to the work-directory.

This should fix #112 in the context of `Biokepi.EDSL`.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/290)


ihodes:
Nice addition, LGTM!

PR #291, opened 2016-07-06
Use better default for `markdup_dups` tmp-dir

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/291)


ihodes:
LGTM!

PR #268, opened 2016-05-12
Support for installing Python tool packages (e.g. vcf-annotate-polyphen)

armish:
This is my first attempt to provide a generalized way to install Python tools from packages/repos. I assume the python module is going to change quite a lot as we move forward with this; but, this should cover the very basic use case. I used [vcf-annotate-polyphen](https://github.com/hammerlab/vcf-annotate-polyphen) for demo purposes here.

I know that there is still a lot to improve here, but I decided to create this PR anyway to collect your early feedback and iterate over these choices as soon as possible. It would be great if you can CR my coding style as well (as I am pretty new to this game).

In the meantime, I might factor some of the conda stuff out to make use of `Python_package` or continue adding Python tools to this PR.

cc: @rleonid @ihodes @smondet

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/268)

PR #292, opened 2016-07-08
Add CombineVariants, AnnotateVariants from GATK

ihodes:
WIP: need to figure out how to best acquire the reference genome for use by these tools…

also 
1. depends on all the VCFs passed into combine_variants
2. make annotate_variants easier to use (right now you really need to write most of what it does into the params/config in strings); this is a confusing tool

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/292)

### DIFF
--- a/src/bfx_tools/gatk.ml
+++ b/src/bfx_tools/gatk.ml
@@ -87,6 +87,14 @@ module Configuration = struct
     include Gatk_config ()
   end
 
+  module Combine_variants = struct
+    include Gatk_config ()
+  end
+
+  module Annotate_variants = struct
+    include Gatk_config ()
+  end
+
   type indel_realigner = (Indel_realigner.t * Realigner_target_creator.t)
   type bqsr = (Bqsr.t * Print_reads.t)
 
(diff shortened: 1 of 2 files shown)


src/bfx_tools/gatk.ml
```diff
@@ -639,3 +647,95 @@ let mutect2
...
+    ?(configuration=Configuration.Combine_variants.default)
+    ?(edges=[])
+    ~run_with
+    ~variant_sources
+    ~output_vcf_path
+    ~genotype_merge_option
+  =
+  let open KEDSL in
+  let reference =
+    Machine.get_reference_genome run_with "" in (* how to get the reference we want to use? *)
```
smondet: pass a `~reference_genome` argument?
Like here: https://github.com/hammerlab/biokepi/blob/c2aac7c1702375485268ae2c1c74f5d9d11a53ff/src/pipeline_edsl/semantics.ml#L188


Later, when this one and #268 are merged, we can create a proper `vcf` workflow-product that (like we do in `bam_file`) carries around its reference-genome.

ihodes:
@smondet back to you

src/bfx_tools/gatk.ml
```diff
@@ -639,3 +647,96 @@ let mutect2
...
+  let gatk = Machine.get_tool run_with Machine.Tool.Default.gatk in
+  let reference_fasta = Reference_genome.fasta reference_genome in
+  let name = sprintf "AnnotateVariants to %s" (Filename.basename output_vcf_path) in
+  let make =
+    Machine.run_big_program run_with ~name
+      ~self_ids:["gatk"; "annotate-variants"]
+      Program.(
+        Machine.Tool.(init gatk)
+        && shf "cd %s" (Filename.dirname output_vcf_path)
+        && call_gatk ~analysis:"CombineVariants" ([
```
smondet: `CombineVariants`?

src/bfx_tools/gatk.ml
```diff
@@ -639,3 +647,96 @@ let mutect2
...
+    ~edges:(edges @ [
+        depends_on Machine.Tool.(ensure gatk);
+        depends_on reference_fasta;
+        on_failure_activate (Remove.file ~run_with output_vcf_path);
+      ])
+
+(** Annotate variant calls with context information.
+
+    Virtually nothing is wrapped in this call; the input files (via the
+    configuration) and edges will need to be manually passed.
```
smondet: I don't understand this, how do you give the input nodes then?

ihodes: You can depend on them via `?edges`, and give the paths via the Configuration record "parameters". The spec of VariantAnnotator is a little much, so I just thought I'd get this in & add to it as I learn what the various modes do/how to use them.

src/bfx_tools/gatk.ml
```diff
@@ -639,3 +647,96 @@ let mutect2
...
+      ~f:(function
+        | `Vcf vcf -> sprintf "--variant %s" vcf#product#path
+        | `Vcf_named_sample (sample_name, vcf) ->
+          sprintf "--variant:%s %s" sample_name vcf#product#path) in
+  let merge_option =
+    "-genotypeMergeOptions " ^
+    match genotype_merge_option with
+    | `Uniquify -> "UNIQUIFY"
+    | `Prioritize samples ->
+      "PRIORITIZE --priority " ^ String.concat ~sep:"," samples
```
iskandr: The priority list isn't required:

>UNION: assumes that each ROD source represents the same set of samples (although this is not enforced). It uses the priority list (if provided) to emit a single record instance at every position represented in the input RODs.

(though I'm not sure what GATK does without it)

Also, shouldn't that be just one dash in `-priority`? 

(In the GATK example it's `-genotypeMergeOptions PRIORITIZE -priority foo,bar`)

ihodes: > The priority list isn't required … (though I'm not sure what GATK does without it)

Me neither, and later in the docs it sounds as though it *is* required, or undefined behavior if it's missing… so I thought I'd just require it here.

> Also, shouldn't that be just one dash in -priority?

Good catch, thank you! GATK requires -- in some places and - in others… lots of fun.

PR #300, opened 2016-07-20
Add `EDSL.mhc_alleles` input function

smondet:
This creates a new functions in the EDSL that is used to input MHC
Alleles for Topiary.

One can give a file name; or input directly allele names.

This is a follow up to #268.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/300)


armish:
Not sure about the feasibility of passing allele names manually in a pipeline, but LGTM! :)

PR #305, opened 2016-07-26
add sample_name to bam_file

arahuja:
This adds `sample_name` as a property of a BAM similar to fastq instead of carrying around a sample name.

Some of the pipeline aspects were a bit confusing, but this should have the same semantics. Are so many definitions of `bam_to_fastq` needed, seemed to have to edit this in a lot of places?

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/305)


smondet:
I forgot we had the one in Bedtools.

All the other ones represent the one in Picard, but implementing `To_json`, `To_workflow`, the older `Pipeline` API, etc.

LGTM will merge soon

PR #308, opened 2016-07-27
Update links to use broad hg38 bundle

arahuja:
The older links were broken, this uses the GATK provided bundle, but this is technically hg38 (meaning the contig are `chr` prefixed.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/308)

### DIFF
--- a/src/run_environment/reference_genome.ml
+++ b/src/run_environment/reference_genome.ml
@@ -27,6 +27,7 @@ module Specification = struct
     metadata: string option;
     fasta: Location.t;
     dbsnp: Location.t option;
+    known_indels: Location.t option;
     cosmic: Location.t option;
     exome_gtf: Location.t option; (* maybe desrves a better name? *)
     cdna: Location.t option;
@@ -40,6 +41,7 @@ module Specification = struct
       ~ensembl
       ~species
       ?dbsnp
+      ?known_indels
       ?cosmic
       ?exome_gtf
       ?cdna
@@ -52,6 +54,7 @@ module Specification = struct
     metadata;
     fasta;
     dbsnp;
+    known_indels;
     cosmic;
     exome_gtf;
     cdna;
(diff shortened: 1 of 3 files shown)


src/run_environment/reference_genome.ml
```diff
@@ -144,11 +144,9 @@ module Default = struct
...
-      "ftp://ftp.ensembl.org/pub/release-79/fasta/homo_sapiens/dna/\
-       Homo_sapiens.GRCh38.dna.primary_assembly.fa.gz" in
+      "ftp://gsapubftp-anonymous@ftp.broadinstitute.org/bundle/hg38/hg38bundle/Homo_sapiens_assembly38.fasta.gz" in
     let dbsnp_b38 =
-      "http://ftp.ncbi.nlm.nih.gov/snp/organisms/human_9606_b142_GRCh38/\
-       VCF/00-All.vcf.gz" in
+      "ftp://gsapubftp-anonymous@ftp.broadinstitute.org/bundle/hg38/hg38bundle/dbsnp_144.hg38.vcf.gz" in
     let gtf_b38_url =
       "http://ftp.ensembl.org/pub/release-79/gtf/homo_sapiens/\
        Homo_sapiens.GRCh38.79.gtf.gz" in
```
smondet: Isn't the GTF going to be using the wrong chronames?

Wouldn't it be better to create a whole new ref-genome `hg38`?

arahuja: Yea that makes sense. I was hoping to avoid that since we will end up with a split - no hg38 GTF, can't find a b38 dbSNP etc.

arahuja:
I updated this to have a separate `h38` reference and:
- removed broken dbsnp link from b38
- added a known indels arg to reference builds as well and added this to hg19, b37 and h38

Issue #29, opened 2015-04-20
Sort and convert sam to bam simultaneously


arahuja:
An unsorted bam is much larger than a sorted one and we can avoid the intermediate file by piping the commands together since almost every tools needs a sorted bam

arahuja:
@ihodes @smondet This might be a way to save some space/time and time on an alignment pipeline? Not writing the original SAM ever would be a large savings of space

smondet:
It might save some time but the `sam` file is already deleted after `sam-to-bam`.

arahuja:
It would save space while running since it would never create the uncompressed output

PR #309, opened 2016-07-27
WIP: Run BWA directly on BAM files

arahuja:
I'm not sure the right way to do this, so opening this for comments. These seems to be a way to get bwa-mem to take BAM as input, but I can't see how to best adjust the pipelines for this as well.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/309)

### DIFF
--- a/src/bfx_tools/bwa.ml
+++ b/src/bfx_tools/bwa.ml
@@ -32,7 +32,6 @@ module Configuration = struct
 
 end
 
-
 let index
     ~reference_build
     ~(run_with : Machine.t) =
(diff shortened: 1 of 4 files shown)


arahuja:
@smondet tried something similar to the first way we discussed, variant of inputs - does this make sense?

smondet:
@arahuja yes, WFM

arahuja:
Closing this - should have looked at BWA further. This will work with `bwa aln` but not `bwa mem` : https://sourceforge.net/p/bio-bwa/mailman/message/30527151/

It would be nice to to pipe some of these items though, would be huge space/time savings I think. For example, `bwa mem | sam2bam | bamsort ` There was another issue on this: https://github.com/hammerlab/biokepi/issues/29

PR #316, opened 2016-08-03
Add igvxml

smondet:
This version works only with `b37decoy` and `mm10` (tool's choice).
But it works™ :)

I didn't add to the TTFI EDSL because I don't see the use case.

Also, the results depends on the paths not on the contents of the files at that path (they can run even if they don't exist), so the tool does not depend on the bam/vcf products.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/316)

### DIFF
--- a/src/bfx_tools/igvxml.ml
+++ b/src/bfx_tools/igvxml.ml
@@ -1,0 +1,47 @@
+
+open Biokepi_run_environment
+open Common
+
+let vcf ~name ~path = object method name = name method path = path end
+
+let run ~(run_with:Machine.t) ~reference_genome
+    ~normal_bam_path
+    ~tumor_bam_path
+    ~run_id
+    ?rna_bam_path
+    ~vcfs
+    ~output_path
+    () =
+  let open KEDSL in
+  let open Ketrew_pure.Target.Volume in
+  let igvxml = Machine.get_tool run_with Machine.Tool.Default.igvxml in
+  let name = sprintf "Create %s" (Filename.basename output_path) in
+  let make =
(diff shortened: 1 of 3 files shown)


smondet:
One funny thing, if called with `b37`, it still silently works with `b37decoy` because Cmdliner compares prefixes for everything.

ihodes:
LGTM. 

Why not add it to the TTFI?

smondet:
@ihodes for now I'm using it **as part of** an extension to the TTFI, but  by itself it doesn't make much sense becase there is no "pipeline" i.e. no dependencies, it's all strings, the API would be

```ocaml
val igvxml:
   normal_bam_path: string ->
   tumor_bam_path: string ->
   ... (* no repr at all *)
   -> [ `Igv_xml ] repr
```

ihodes:
Ah, I see what you mean. The downside to this is as a consumer of the TTFI, you'd need to write a fair amount of code to be able to use this in your own pipeline, right?

smondet:
Even in the TTFI, as is, it would be pretty awkward to use. You need to give paths that depend on the backend actually running (in our case choices made in `Biokepi.Tools`) and on top of that you may have to transform those paths to something else (in my case relative paths from the place the results are save, not computed; but it may also be that you have to transform into HTTP URLs (?)).
Within a TTFI pipeline those considerations are too low-level, aren't they?

ihodes:
Hmm, not really sure—the alternative is cumbersome. Regardless, this looks good, and thank you for adding it!

PR #317, opened 2016-08-04
Add hlarp

ihodes:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/317)

### DIFF
--- a/src/pipeline_edsl/semantics.ml
+++ b/src/pipeline_edsl/semantics.ml
@@ -130,6 +130,10 @@ module type Bioinformatics_base = sig
     [ `Fastq ] repr ->
     [ `Optitype_result ] repr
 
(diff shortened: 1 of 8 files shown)


src/pipeline_edsl/semantics.ml
```diff
@@ -130,6 +130,10 @@ module type Bioinformatics_base = sig
     [ `Fastq ] repr ->
     [ `Optitype_result ] repr
 
+  val hlarp:
+    [`Optitype_result | `Seq2hla_result] repr ->
```
smondet: So `_ repr` is not covariant this function AFAIK cannot be called.

Discuss at #318 :)

ihodes:
this works!

PR #319, opened 2016-08-05
Use the version string in the Conda environment

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/319)

### DIFF
--- a/src/environment_setup/python_package.ml
+++ b/src/environment_setup/python_package.ml
@@ -29,7 +29,8 @@ let create_python_tool ~host ~(run_program : Machine.Make_fun.t) ~install_path
   in
   let main_subdir = name ^ "_conda_dir" in
   let conda_env = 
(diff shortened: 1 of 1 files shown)


ihodes:
LGTM

might be a good addition for Conda.setup_env?

smondet:
@ihodes what do you mean?

ihodes:
`Conda.setup_environment ?version …` instead of manually adding it to the name?

smondet:
@ihodes but the `version` is the version of the package. We happen to create one Conda env per package because it doesn't really work to share them; but from Conda's point of view it's just an environment with a name.
Cf. the `Biopam` module which has a somewhat different use of `Conda`:
https://github.com/hammerlab/biokepi/blob/master/src/environment_setup/biopam.ml#L200
(no `version` there).

PR #323, opened 2016-08-09
Add mutect2 config without cosmic

arahuja:
Still not sure this should be an option, but right now hoping to get a pipeline moving.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/323)


ihodes:
LGTM!

PR #339, opened 2016-08-22
Fix bedtools bamtofastq

arahuja:
Looks this actually drops the dependency on the sorted bam and just uses original input bam.

### DIFF
--- a/src/bfx_tools/bedtools.ml
+++ b/src/bfx_tools/bedtools.ml
@@ -43,7 +43,7 @@ let bamtofastq
       (["-fq"; r1], r1, None)
   in
   let bedtools = Machine.get_tool run_with Machine.Tool.Default.bedtools in
-  let src_bam = input_bam#product#path in
+  let src_bam = sorted_bam#product#path in
   let program =
     Program.(Machine.Tool.(init bedtools)
              && exec ["mkdir"; "-p"; Filename.dirname r1]
@@ -57,6 +57,7 @@ let bamtofastq
   let edges = [
     depends_on Machine.Tool.(ensure bedtools);
     depends_on input_bam;
+    depends_on sorted_bam;
     on_failure_activate (Remove.file ~run_with r1);
     on_success_activate (Remove.file ~run_with sorted_bam#product#path);
   ] |> fun list ->


src/bfx_tools/bedtools.ml
```diff
@@ -43,7 +43,7 @@ let bamtofastq
       (["-fq"; r1], r1, None)
   in
   let bedtools = Machine.get_tool run_with Machine.Tool.Default.bedtools in
-  let src_bam = input_bam#product#path in
+  let src_bam = sorted_bam#product#path in
```
smondet: Can you also remove the `sorted_bam` value also?
(github does not let me comment on the non-diff line above)

smondet:
So what does `bedtools bamtofastq` do on non-sorted bams?
it outputs the whole FASTQ in "aligned order"?

src/bfx_tools/bedtools.ml
arahuja: What do you mean remove it?  L31?

arahuja:
> So what does bedtools bamtofastq do on non-sorted bams?

It gives a warning on every single read that it is not next to its mate and then skips the read.

src/bfx_tools/bedtools.ml
smondet: Yes it seems that it should give you an "unused value" warning, doesn't it?

smondet:
> It gives a warning on every single read that it is not next to its mate and then skips the read.

so if the alignment  is like `R1a R1b R2a R2b` both a and b pairs will be dropped?

src/bfx_tools/bedtools.ml
arahuja: What do you mean? `sorted_bam` is used in L46 now, before it was unused.

arahuja:
> so if the alignment is like R1a R1b R2a R2b both a and b pairs will be dropped?
In this case both would be kept (if I am reading it correctly), because coordinate order matches name order

src/bfx_tools/bedtools.ml
smondet: I completely read the diff backwards!
`OutOfCoffeeException`!
I understand now :)

arahuja:
Cool thanks! For clarification for the future - is the `depends_on sorted_bam` needed? Or does the fact that this command takes an input a workflow implicitly add that workflow to `depends_on`?

smondet:
@arahuja no, there is nothing implicit.
Here the redundant one is `depends_on input_bam` because the sorted bam already depends on `input_bam`.

PR #350, opened 2016-09-09
Control filename lengths

smondet:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/350)

### DIFF
--- a/src/run_environment/common.ml
+++ b/src/run_environment/common.ml
@@ -84,17 +84,17 @@ module Name_file = struct
     let max_length = 220 in
     let buf = Buffer.create max_length in
     Buffer.add_string buf hash;
-    let rec loop =
+    let rec append_components =
       function
       | [] -> ()
       | one :: more ->
         if
           Buffer.length buf + String.length readable_suffix
           + String.length one < max_length
-        then (Buffer.add_string buf one; loop more)
+        then (Buffer.add_string buf one; append_components more)
         else ()
     in
-    loop components;
+    append_components components;
     Buffer.add_string buf readable_suffix;
     let name = Buffer.contents buf in
     begin if String.length name > max_length then
--- a/src/run_environment/common.ml
+++ b/src/run_environment/common.ml
@@ -46,6 +46,86 @@ module Unique_id = struct
   include Ketrew_pure.Internal_pervasives.Unique_id
 end
 
+(** 
+
+Generate unique filenames for a given set of uniquely identifying properties.
+
+The module generates names that remain (way) under 255 bytes because they
+are the most common
+{{:https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits}file-system limits}
+in 2016.
+
+*)
+module Name_file = struct
+  (* Additional safety: we check that there are no hash-duplicates
+     within a given execution of a program linking with Biokepi. *)
+  let db : (string, string list) Hashtbl.t = Hashtbl.create 42
+
+  let path ~readable_suffix ~from high_level_components =
+    let sanitize =
+      String.map
+        ~f:(function
+          | ('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '-') as c -> c
+          | other -> '_') in
+    let components =
+      begin match from with
+      | `Path p ->
+        let b = Filename.basename p in
+        (try [Filename.chop_extension b] with _ -> [b])
+      | `In_dir d -> []
+      end
+      @
+      List.map high_level_components ~f:sanitize
+    in
+    let hash =
+      String.concat ~sep:"-" (readable_suffix :: components)
+      |> Digest.string |> Digest.to_hex
+    in
+    let max_length = 220 in
+    let buf = Buffer.create max_length in
+    Buffer.add_string buf hash;
+    let rec loop =
+      function
+      | [] -> ()
+      | one :: more ->
+        if
+          Buffer.length buf + String.length readable_suffix
+          + String.length one < max_length
+        then (Buffer.add_string buf one; loop more)
+        else ()
+    in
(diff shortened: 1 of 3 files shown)


src/run_environment/common.ml
```diff
@@ -46,6 +46,86 @@ module Unique_id = struct
...
+      | `In_dir d -> []
+      end
+      @
+      List.map high_level_components ~f:sanitize
+    in
+    let hash =
+      String.concat ~sep:"-" (readable_suffix :: components)
+      |> Digest.string |> Digest.to_hex
+    in
+    let max_length = 220 in
```
ihodes: why not 255?

src/run_environment/common.ml
```diff
@@ -46,6 +46,86 @@ module Unique_id = struct
...
+    let buf = Buffer.create max_length in
+    Buffer.add_string buf hash;
+    let rec loop =
```
ihodes: A more descriptive name would be nice, like "construct_name" or something?

src/run_environment/common.ml
```diff
@@ -46,6 +46,86 @@ module Unique_id = struct
...
+          Buffer.length buf + String.length readable_suffix
+          + String.length one < max_length
+        then (Buffer.add_string buf one; loop more)
+        else ()
+    in
+    loop components;
+    Buffer.add_string buf readable_suffix;
+    let name = Buffer.contents buf in
+    begin if String.length name > max_length then
+        ksprintf failwith "Trying to make filename too long: %s" name
```
ihodes: "Filename too long (max 220 (or 255?)): %s"

src/run_environment/common.ml
```diff
@@ -46,6 +46,86 @@ module Unique_id = struct
...
+    let sanitize =
+      String.map
+        ~f:(function
+          | ('a' .. 'z' | 'A' .. 'Z' | '0' .. '9' | '-') as c -> c
+          | other -> '_') in
+    let components =
+      begin match from with
+      | `Path p ->
+        let b = Filename.basename p in
+        (try [Filename.chop_extension b] with _ -> [b])
```
ihodes: Why remove the extension?

src/run_environment/common.ml
```diff
@@ -46,6 +46,86 @@ module Unique_id = struct
...
+{{:https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits}file-system limits}
+in 2016.
+
+*)
+module Name_file = struct
+  (* Additional safety: we check that there are no hash-duplicates
+     within a given execution of a program linking with Biokepi. *)
+  let db : (string, string list) Hashtbl.t = Hashtbl.create 42
+
+  let path ~readable_suffix ~from high_level_components =
```
ihodes: Why have `~from` at all; could it all just go in the `high_level_components` section (which might be more simply named `components`?

ihodes:
Couple comments, LGTM overall; back to you!

Overall Q: how does a user figure out which of two `HASHBLAH-mutect-default.vcf" comes from, where one was e.g. indel realigned and one wasn't?

src/run_environment/common.ml
smondet: We need a margin, since some tools have the *“algomerating naming*” encoded; e.g. bam indexes have to always be `<bam-file>.bai`.

The function is also used for many `~result_prefix` arguments that will add something at the end (and use it also for temporary/intermediate files).

src/run_environment/common.ml
smondet: we don't want extensions in the middle of filenames (?)

The function packs as many components as possible in the filename for readability/debuggability.

src/run_environment/common.ml
smondet: I started like that, and seeing how the function is called made me make the distinction, it's just easier/nicer to call like that.

smondet:
> Overall Q: how does a user figure out which of two `HASHBLAH-mutect-default.vcf" comes from, where one was e.g. indel realigned and one wasn't?

Before this PR the "aggregate" naming was used only for quick debugging.

The user should always track the results their interested in (that why they're always available in node "products" or, when EDSLing, in `To_workflow.File_specification.t`).

One day, we could dump offer more facilities to dump the `Hashtbl.t` to a file if needed.

src/run_environment/common.ml
smondet: 👍

src/run_environment/common.ml
smondet: 👍

src/run_environment/common.ml
ihodes: I'm not sure I understand? It seems a little unusual see a Name_file.from_path <one of 4 paths that are input into this one, arbitrarily chosen> [the other paths and config etc]

src/run_environment/common.ml
ihodes: Ah right—might be worth noting this length in the documentation, then, since it's important for callers to know. (Could also be nice to make it an option, but not a big deal!)

src/run_environment/common.ml
smondet: `from_path` means: take a path and derive a new similar one (i.e. same directory, and "depend" on it).

since many tools take one file and convert it to something else, it's just easier (no need to `Filename.basename`, `Filename.chop_extension`, `Filename.dirname` every time).

screenshot 2016-09-09 14 12 15

If the need really arises we can add `Name_file.from_paths` that takes more than one path (didn't seem that common so far).

ihodes: Ah yes, that makes sense; I wasn't thinking of the directory portion of the equation.

ihodes:
LGTM

PR #352, opened 2016-09-13
Fix module paths in `EDSL.To_workflow`

smondet:
Without the module paths, the interaction between packed modules and cmis goes nuts (we've seen that before).

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/352)


ihodes:
LGTM

PR #351, opened 2016-09-12
add gawk as a dependency (for NetMHC stuff)

armish:
Yet another follow up to #347 and #287.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/351)

Issue #287, opened 2016-06-15
Add support for MHC binding prediction tools


armish:
#268 adds topiary (#142) to our toolset, but it still doesn't address installing the NetMHC tool collection. Given that the distribution of those tools is somewhat indirect, we need a solution that is similar to what we have for Broad's JAR-based tools.

PR #353, opened 2016-09-14
Better tame the NetMHC tool installations

armish:
follow-up to #287, #347, and #351 to fix various issues with the latest setup:

- Installs a clean Python 2 environment to be used by NetMHC tools (so that no other Python installation interferes with their function)
  - Also fixes the old NetMHC script to force it use whatever Python is available in the `$PATH`
- Improves `Conda`'s init function in such a way that if the environment is already active, it won't get activated once more (provides smooth handling of env variables if one is to init multiple netMHC tools independently)
- Now cleans up after itself on failure
- Creates the `$TMPDIR` beforehand to satisfy `NetMHC`'s weak assumptions

All of these tested on a clean VM on Gcloud, so this should be good to go.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/353)

### DIFF
--- a/src/environment_setup/netmhc.ml
+++ b/src/environment_setup/netmhc.ml
@@ -189,7 +189,8 @@ let default_netmhc_install
           shf "tar zxf %s" downloaded_file#product#path &&
           shf "cd %s" tool_path &&
           chain (List.map ~f:fix_script env_setup) &&
-          shf "chmod +x %s" binary_path
+          shf "chmod +x %s" binary_path &&
+          shf "mkdir -p %s" (tmp_dir install_path)
         )
       )
   in
--- a/src/environment_setup/netmhc.ml
+++ b/src/environment_setup/netmhc.ml
@@ -6,7 +6,7 @@ open Common
 
     netMHC-4.0a.Linux.tar.gz // netMHC-3.4a.Linux.tar.gz
     pickpocket-1.1a.Linux.tar.gz
-    netMHCpan-3.0a.Linux.tar.gz
+    netMHCpan-3.0a.Linux.tar.gz // netMHCpan-2.8a.Linux.tar.gz
     netMHCcons-1.1a.Linux.tar.gz
 
   Do not use custom named archives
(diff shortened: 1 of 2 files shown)


src/environment_setup/netmhc.ml
```diff
@@ -40,11 +42,16 @@ let replace_env_value file envname newvalue =
   let file_bak = file_org ^ ".bak" in
   KEDSL.Program.(
     shf "mv %s %s" file_org file_bak &&
-    shf "sed -e 's/setenv\t%s\t.*/setenv\t%s\t%s/g' %s > %s"
-      envname envname (escape_slash newvalue) file_bak file_org &&
+    shf "sed -e 's/%s/%s/g' %s > %s"
+      (escape_slash oldvalue) (escape_slash newvalue) file_bak file_org &&
     shf "rm -f %s" file_bak
```
smondet: Wow! Editing tcsh scripts with sed !
👏

src/environment_setup/netmhc.ml
```diff
@@ -196,19 +227,24 @@ let default ~run_program ~host ~install_path ~(files:netmhc_file_locations) () =
...
     | None -> true
   in
+  let netmhc_env = guess_env_setup ~install_path files.netmhc in
   let older_netmhc =
     default_netmhc_install ~run_program ~host ~install_path
     ~tool_file_loc:files.netmhc ~binary_name:"netMHC"
     ~example_data_file:(Some "SLA-10401/bl50/synlist") 
-    ~env_setup:(guess_env_setup ~install_path files.netmhc)
+    ~env_setup:(
+      [ `GENERIC ("/usr/local/bin/python2.5", "`which python`") ]
```
smondet: Not sure I understand, at which point are the backquotes being executed?
In the tcsh script?

src/environment_setup/netmhc.ml
armish: haters gonna hate 🚶 ;)

src/environment_setup/netmhc.ml
armish: yeah, in the tcsh script. We could just replace it with the full path to the python that we get from the Conda, but this comes with some flexibility. Also: in the file where the script decides on what python to use, the other option is this one (so this is as it is for consistency).

smondet: :thumbs_up:

PR #355, opened 2016-09-15
NetMHC: wrap the netMHC wrappers for better environment control

armish:
so that when another tool (e.g. vaxrank) that depends on another version of Python (e.g. v3) calls one of these scripts, we make sure that we first enter into the environment we set up (Python v2), run the tool and go back to the whatever environment we had :man::gun:

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/biokepi/355)


smondet:
LGTM!
Waiting on @ihodes ' stamp :)

ihodes:
LGTM! This is amazing. 

I also think you get the Hammer Lab Centennial Award for Persevering in the Face of Insanity and Absurdity.

PR #358, opened 2016-09-19
Vaxrank depends on the input BAM having a BAI

### LINKED ISSUE #356: Add dependence on .bai of bam file for Vaxrank

ihodes:
fixes #356

Question: will this fail if the BAM is unsorted? Does Vaxrank expect the BAM to be sorted? @iskandr

### DIFF
--- a/src/bfx_tools/vaxrank.ml
+++ b/src/bfx_tools/vaxrank.ml
@@ -172,12 +172,13 @@ let run ~(run_with: Machine.t)
     product
     ~name
     ~edges:([
-      depends_on Machine.Tool.(ensure vaxrank);
-      depends_on (Pyensembl.cache_genome ~run_with ~reference_build);
-      depends_on bam;
-      depends_on alleles_file;
-    ] @ (List.map ~f:depends_on vcfs)
-      @ predictor_edges)
+        depends_on (Samtools.index_to_bai ~run_with bam);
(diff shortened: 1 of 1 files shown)


arahuja:
@ihodes IIRC you shouldn't be able to index an unsorted BAM

ihodes:
@arahuja Thanks! cool, that's what I thought, will add that in.

src/bfx_tools/vaxrank.ml
```diff
@@ -172,12 +174,13 @@ let run ~(run_with: Machine.t)
...
+        depends_on (Samtools.index_to_bai ~run_with sorted_bam);
+        depends_on Machine.Tool.(ensure vaxrank);
+        depends_on (Pyensembl.cache_genome ~run_with ~reference_build);
+        depends_on bam;
+        depends_on alleles_file;
+      ] @ (List.map ~f:depends_on vcfs)
+        @ predictor_edges)
     ~make:(
       Machine.run_program run_with ~name
         Program.(
```
smondet: You also need to use `sorted_bam` in the `make` also (cf. `let arguments` above)

ihodes: Good catch, thank you

ihodes:
@smondet fixed, back to you

PR #359, opened 2016-09-22
Implement vcf product

smondet:
As we discussed once, there is now a `vcf_file` ketrew-product.

This makes some `reference_build` arguments redundant/useless in the EDSL, so the API-breaking change is that simplification (minor PR to epidisco on the way).

While propagating I fixed a minor bug in `isovar`'s `To_json`/`To_display`.

I also added a couple more "conformity checks" (that VCFs and Bams have the same reference-build).

### DIFF
--- a/src/run_environment/common.ml
+++ b/src/run_environment/common.ml
@@ -295,6 +295,25 @@ module KEDSL = struct
     | Single_bam: bam_file workflow_node -> bam_file workflow_node bam_or_bams
     | Bam_workflow_list: bam_file workflow_node list -> bam_list workflow_node bam_or_bams
 
+  type vcf_file = <
+    is_done: Ketrew_pure.Target.Condition.t option;
+    host: Host.t;
+    path : string;
+    reference_build: string;
+    as_single_file: single_file product;
+  >
+  let vcf_file ~host ~reference_build path : vcf_file =
+    object (self)
+      val file = single_file ~host path
+      method host = host
+      method path = file#path
+      method is_done = file#is_done
+      method reference_build = reference_build
+      method as_single_file = file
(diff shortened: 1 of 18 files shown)


src/run_environment/common.ml
```diff
@@ -295,6 +295,25 @@ module KEDSL = struct
...
+    object (self)
+      val file = single_file ~host path
+      method host = host
+      method path = file#path
+      method is_done = file#is_done
+      method reference_build = reference_build
+      method as_single_file = file
+    end
+  let transform_vcf vcf ~path =
+    vcf_file ~host:vcf#host ~reference_build:vcf#reference_build vcf#path
```
ihodes: need to use the path

src/run_environment/workflow_utilities.ml
```diff
@@ -353,10 +356,12 @@ module Vcftools = struct
       ~vcftools
       ~(run_program : Machine.Make_fun.t)
       ?more_edges
+      ~make_product
       ~src ~dest () =
     let run_program =
       Machine.Make_fun.with_requirements run_program [`Memory `Big] in 
     vcf_process_n_to_1_no_machine
+      ~make_product:(fun p -> KEDSL.single_file p ~host)
```
ihodes: not a VCF output?

src/run_environment/common.ml
smondet: Fixed, thanks :)

src/run_environment/workflow_utilities.ml
smondet: Oups, copy-paste error. It should be just `~make_product`

PR #371, opened 2016-10-10
Python-tools: bump vaxrank version

armish:
@ihodes I remember you fiddling with this. Any reason why we haven't done this yet?

cc: @iskandr @julia326

### DIFF
--- a/src/environment_setup/python_package.ml
+++ b/src/environment_setup/python_package.ml
@@ -72,5 +72,5 @@ let default ~host ~run_program ~install_path () =
     create_python_tool ~host ~run_program ~install_path
       ~version:"0.0.21" (Pip, Package_PyPI "topiary");
     create_python_tool ~host ~run_program ~install_path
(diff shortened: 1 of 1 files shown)


src/environment_setup/python_package.ml
```diff
@@ -72,5 +72,5 @@ let default ~host ~run_program ~install_path () =
     create_python_tool ~host ~run_program ~install_path
       ~version:"0.0.21" (Pip, Package_PyPI "topiary");
     create_python_tool ~host ~run_program ~install_path
-      ~version:"0.2.0" (Pip, Package_PyPI "vaxrank");
+      ~version:"0.2.5" (Pip, Package_PyPI "vaxrank");
```
smondet: This list defines what biokepi will pickup by default (with no cofig) when a user tries to use vaxrank.

So when a new vaxrank is realeased the question is "do we want it as the ‘default’".

I think for minor version bumps, yes, but for bigger changes we need to test and understand what they mean for the pipelines.

PR #376, opened 2016-10-19
Fix the naming hash comparing function types

ihodes:
Also rename the suffix for bedtools intersect in the TTFI

### DIFF
--- a/src/run_environment/common.ml
+++ b/src/run_environment/common.ml
@@ -102,7 +102,9 @@ module Name_file = struct
           name max_length
     end;
     begin match Hashtbl.find db name with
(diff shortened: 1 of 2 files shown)


src/run_environment/common.ml
```diff
@@ -102,7 +102,9 @@ module Name_file = struct
           name max_length
     end;
     begin match Hashtbl.find db name with
-    | some when List.sort some = List.sort components -> ()
+    | some
+      when List.sort ~cmp:String.compare some
+           = List.sort ~cmp:String.compare components -> ()
```
smondet: Oh yes, the `~cmp` is not optional !

PR #377, opened 2016-10-19
Simplify EDSL extensions

smondet:


ihodes:
LGTM! Yay for simplification!

PR #382, opened 2016-11-16
fix OptiType installation and init procedure

armish:
These two fixes are needed to get the OptiType working again:

- the version postfix in the folder name generation was breaking how we built the path to the OptiType root
- glpk package that OptiType depends on is not always available with default distributions and even if it is, the relatively new versions do not play nicely with the OptiType. This fixes the version of the glpk package to the one that is compatible with OptiType (more details here: https://github.com/FRED-2/OptiType/issues/28)

Just realized that I completely forgot to push this fix out and make a PR off of it :\

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,4 @@ os:
 - osx
 sudo: true
 env:
-  - OCAML_VERSION=4.02.3 OPAM_VERSION=1.2.1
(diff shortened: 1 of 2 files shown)


smondet:
The Travis failure is because the script pulls Ketrew's `master` branch that has dropped support for OCaml 4.02.x.

ihodes:
LGTM! 

Mind removing the 4.02 tests from the travis.yml?

PR #386, opened 2016-12-02
Updating Dockerfile to allow for vaxrank PDF report generation:

julia326:
- adding non-Python dependencies
- adding a font config file needed for good kerning in output PDF (Ubuntu issue)

Tested by creating a local Docker image with this Dockerfile and running vaxrank.

### DIFF
--- a/tools/docker/run/Dockerfile
+++ b/tools/docker/run/Dockerfile
@@ -5,10 +5,15 @@
 FROM hammerlab/ketrew-server
 # Installing easy Biokepi dependencies:
 RUN sudo apt-get update
-RUN sudo apt-get install --yes cmake r-base tcsh libx11-dev libfreetype6-dev pkg-config wget gawk graphviz
+RUN sudo apt-get install --yes cmake r-base tcsh libx11-dev libfreetype6-dev pkg-config wget gawk graphviz xvfb
+
+# install wkhtmltopdf from source, this version comes with patched QT necessary for PDF gen
+RUN cd /tmp ; wget http://download.gna.org/wkhtmltopdf/0.12/0.12.3/wkhtmltox-0.12.3_linux-generic-amd64.tar.xz
+RUN cd /tmp && tar xvfJ wkhtmltox-0.12.3_linux-generic-amd64.tar.xz
+RUN cd /tmp/wkhtmltox/bin && sudo chown root:root wkhtmltopdf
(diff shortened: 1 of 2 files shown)


tools/docker/run/Dockerfile
```diff
@@ -19,11 +24,14 @@ RUN sudo bash -c "DEBIAN_FRONTEND=noninteractive apt-get install --yes --allow-u
...
   adduser --uid 20042 --disabled-password --gecos '' biokepi && \
   passwd -l biokepi && \
   chown -R biokepi:biokepi /home/biokepi"
 USER biokepi
 ENV HOME /home/biokepi
 WORKDIR /home/biokepi
 RUN opam init -a --yes --comp 4.02.3
+
+COPY fonts.conf $HOME/.config/fontconfig/fonts.conf
+RUN ln -s $HOME/.config/fontconfig/fonts.conf $HOME/.fonts.conf
```
smondet: (Just out of curiosity)  is there a know way to make this setup system-wide?

That would make it a bit less fragile (?).

julia326: System-wide meaning outside of Linux? Annoyingly, not really - HTML/PDF rendering is proving to cause platform-specific problems, so this solution is only necessary on Ubuntu. But it shouldn't break other things. What other platforms do we run biokepi on?

smondet: No I meant for the whole system as opposed to for the current user (`$HOME` is `/home/biokepi`).

smondet:
The Travis error seems to be (again) because of the new version of `omake`.

tools/docker/run/Dockerfile
```diff
@@ -19,11 +24,14 @@ RUN sudo bash -c "DEBIAN_FRONTEND=noninteractive apt-get install --yes --allow-u
...
   adduser --uid 20042 --disabled-password --gecos '' biokepi && \
   passwd -l biokepi && \
   chown -R biokepi:biokepi /home/biokepi"
 USER biokepi
 ENV HOME /home/biokepi
 WORKDIR /home/biokepi
 RUN opam init -a --yes --comp 4.02.3
+
+# Copy local fonts config over, also needed for PDF gen
+COPY fonts.conf /etc/fonts/local.conf
```
smondet: Cool :+1: good to know :)

julia326:
Partially fixes https://github.com/hammerlab/epidisco/issues/93

PR #393, opened 2016-12-14
Add Sambamba

### LINKED ISSUE #392: Add sambamba
https://github.com/lomereiter/sambamba/wiki/%5Bsambamba-view%5D-Filter-expression-syntax

ihodes:
Fixes #392

### DIFF
--- a/src/bfx_tools/sambamba.ml
+++ b/src/bfx_tools/sambamba.ml
@@ -4,6 +4,8 @@ open Common
 module Remove = Workflow_utilities.Remove
 
 
+(* Filter language syntax at
+   https://github.com/lomereiter/sambamba/wiki/%5Bsambamba-view%5D-Filter-expression-syntax *)
 let view ~(run_with : Machine.t) ~bam ~filter output_file_path =
   let open KEDSL in
   let name =
--- a/src/bfx_tools/sambamba.ml
+++ b/src/bfx_tools/sambamba.ml
@@ -3,6 +3,27 @@ open Common
 
 module Remove = Workflow_utilities.Remove
 
+module Filter = struct
+  type t = [
+    `String of string
+  ]
+
+  let of_string s =
+    `String s
+
+  let to_string f =
+    match f with
+    | `String s -> s
+
+  module Defaults = struct
+    let only_split_reads =
+      of_string "cigar =~ /^.*N.*$/"
+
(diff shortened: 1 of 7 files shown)


src/bfx_tools/sambamba.ml
```diff
@@ -0,0 +1,36 @@
+open Biokepi_run_environment
+open Common
+
+module Remove = Workflow_utilities.Remove
+
+
+(* Filter language syntax at
+   https://github.com/lomereiter/sambamba/wiki/%5Bsambamba-view%5D-Filter-expression-syntax *)
+let view ~(run_with : Machine.t) ~bam ~filter output_file_path =
```
smondet: I would create a new type for filters:

```ocaml
type filter = [
| `Raw of string
]
```

so that later, if useful, we can add a better-typed alternative  (like `` `Expression of Ast.t``).

src/run_environment/machine.ml
```diff
@@ -33,6 +33,7 @@ module Tool = struct
     open Definition
     let bwa = create "bwa" ~version:"0.7.10"
     let freebayes = create "freebayes" ~version:"1.1.0"
+    let sambamba = create "sambamba" ~version:"v0.6.5"
```
smondet: Why the `"v"` prefix?

src/pipeline_edsl/semantics.ml
```diff
@@ -180,6 +180,14 @@ module type Bioinformatics_base = sig
     [ `Bam ] repr ->
     [ `Bam ] repr
 
+  (** Sambamba's view filter tool, used to filter down a BAM to one with reads
+      matching some predicate (filter language at https://github.com/lomereiter/sambamba/wiki/%5Bsambamba-view%5D-Filter-expression-syntax).
+  *)
+  val filter:
```
smondet: I would at least call it `filter_bam` but I even prefer `sambamba_filter`.

src/run_environment/machine.ml
ihodes: No need; that's just what they called their version, but I'll drop it here.

src/bfx_tools/sambamba.ml
ihodes: SGTM, was thinking about it but thought that'd be an easy addition if the day comes.

PR #407, opened 2017-01-19
Update vaxrank to 0.4.0

ihodes:


src/environment_setup/python_package.ml
```diff
@@ -72,5 +72,5 @@ let default ~host ~run_program ~install_path () =
     create_python_tool ~host ~run_program ~install_path
       ~version:"0.0.21" (Pip, Package_PyPI "topiary");
     create_python_tool ~host ~run_program ~install_path
-      ~version:"0.3.1" (Pip, Package_PyPI "vaxrank");
+      ~version:"0.4.0" (Pip, Package_PyPI "vaxrank");
```
julia326: Do we also want to enable filtering by mismatches after the variant, by default?

ihodes: I don't think so; isn't that for use with FFPE/RiboZero prepped samples? /cc @iskandr

julia326: My understanding was that we want to do this regardless, it'll just have more of an effect with FFPE/RiboZero samples. Was it an FFPE sample that initially motivated this change?

ihodes:
Added a new commit exposing new param directly.

src/environment_setup/python_package.ml
ihodes: I'm not sure what the reasoning was; looking for @iskandr input here. Why didn't we do this before for e.g.? If it's not a default in Vaxrank (since it's by default off), I'd lean towards leaving it that way in general (what's actually run in Epidisco is a different matter, though; we can default to whatever we'd like there, I'm just not sure what the reasoning is for going one way over another).

iskandr: I thought we were just going to turn this on for FFPE samples to allow for the remote possibility of detecting alternate splicing after the variant. Since we haven't yet actually seen this it must be at least moderately rare for a variant to cause downstream splicing which still aligns well enough for us to detect.

PR #424, opened 2017-02-13
Make biokepi use master branch igvxml

ihodes:


smondet:
I'm not sure it's a good idea to use for the "defaults."

You can force the `master` branch to be used in your biokepi machine.

ihodes:
Should I just use a commit in this case (to use a newer version of igvxml?)

hammer:
@ihodes why not just cut a release of igvxml?

ihodes:
Because updating on opam is kind of a pain for no benefit right now, and I anticipate continuing to change igvxml (lots of features need to be added) quite a bit in the near future for other uses of it.

hammer:
Now that we have real patients in the pipeline I'd prefer we rely on released versions of software that can be recorded to improve reproducibility.

I also feel like a lot of pain for new users is caused by pinning.

ihodes:
Gotcha—this is totally separate from the user needing to pin anything or know about what's happening. Biokepi pins to a commit here, and the commit hash would be noted so it's entirely reproducible.

smondet:
@ihodes 

> I anticipate continuing to change igvxml (lots of features need to be added)

for that the best is using your `Biokepi.Machine.t` to force the one you want.

-----

I agree with @hammer the defaults should be proper releases (with `opam-publish` the release process is pretty fast also).

ihodes:
> for that the best is using your Biokepi.Machine.t to force the one you want.

This is something I want all users to have, since it's effectively broken. 

opam-publish isn't fast: it requires manual review. 

Is there any reason to not be okay using a git tag or commit? As far as I can tell, it 100% addresses any concerns either of you have?

smondet:
> This is something I want all users to have…

That's why it's called "Release" ;)

> 
> opam-publish isn't fast: it requires manual review.

Well fully-automated could only be wrong…

> 
> Is there any reason to not be okay using a git tag or commit? As far as I can
> tell, it 100% addresses any concerns either of you have?

A github-only release (glorified tag) works for me.
A bit sad to not push to the mothership opam-repo.

ihodes:
Cool, I'll push a tag out for this commit.

hammer:
Please also try to release regularly to OPAM as well.

ihodes:
@smondet back to you!

PR #425, opened 2017-02-14
Use our own (ordered) version of b38

ihodes:


smondet:
Can you comment (here) how you produced the file? for the historians.

ihodes:
I ran fasta-orderer (from ogene) on b38.fasta (similarly with mm10).

PR #428, opened 2017-02-15
Edsl.concat: fix and improve

smondet:
Tested both commits separately, the first one fixes issues with file-naming, and the second one just avoids useless workflow-nodes that do `cat file.fastq > file-concat.fastq`.


ihodes:
LGTM! Thanks for fixing.

PR #429, opened 2017-02-15
Add BAM to BAI to the TTFI

tavinathanson:
Easily convert BAMs to BAIs in a pipeline.


tavinathanson:
I think @ihodes recommended `bai`, I'll defer to you both since I'm less familiar with the words used around this code base.

ihodes:
*discussed offline, tl;dr function will be a verb, type remains Bai*

tavinathanson:
Renamed function to `index_bam`.

PR #430, opened 2017-02-16
Bump pyensembl version to make it compatible with the latest versions of pygenomics tools

armish:
This new version comes with data structures and filenames for the cached data, so is needed for compatibility across all the pygenomics tools (including vaxrank, isovar, topiary, ...). This and `vaxrank` apparently became out-of-sync after #407 and vaxrank fails when the cache is coming from an older version of pyensembl.


ihodes:
LGTM

PR #431, opened 2017-02-23
Add fastq_of_input case for Input.t.Bam

ihodes:
Also rename Input.of_bam to fastq_of_bam

Improve error messages

this fixes an issue in epidisco with the no-realign option @julia326


smondet:
The Travis failure is actually a real issue \o/

```ocaml
File "./src/examples/edsl_extension_register_result.ml", line 333, characters 6-12:
Error: Unbound value of_bam
```

PR #432, opened 2017-02-25
Bump topiary version and add multiple VCF support

armish:
Depends on https://github.com/hammerlab/topiary/pull/67.

With this, we will be able to pass multiple VCFs to topiary and this will make it easier to include topiary in the pipeline (cf. https://github.com/hammerlab/epidisco/issues/158)

### DIFF
--- a/src/pipeline_edsl/to_workflow.ml
+++ b/src/pipeline_edsl/to_workflow.ml
@@ -859,8 +859,7 @@ module Make (Config : Compiler_configuration)
       vcfs predictor alleles =
     let vs = List.map ~f:get_vcf vcfs in
     let refs = 
-      List.map ~f:(fun v -> v#product#reference_build)
-      |> List.dedup
+      vs |> List.map ~f:(fun v -> v#product#reference_build) |> List.dedup
     in
     let reference_build =
       if List.length refs > 1
--- a/src/pipeline_edsl/to_workflow.ml
+++ b/src/pipeline_edsl/to_workflow.ml
@@ -858,15 +858,17 @@ module Make (Config : Compiler_configuration)
   let topiary ?(configuration=Tools.Topiary.Configuration.default)
       vcfs predictor alleles =
     let vs = List.map ~f:get_vcf vcfs in
-    let refs = List.map ~f:(fun v -> v#product#reference_build) vs in
-    let oneref = List.nth_exn refs 0 in
+    let refs = 
+      List.map ~f:(fun v -> v#product#reference_build)
+      |> List.dedup
+    in
     let reference_build =
-      if List.exists refs ~f:(fun r -> r <> oneref)
+      if List.length refs > 1
       then
         ksprintf failwith "VCFs do not agree on their reference build: %s"
           (String.concat ~sep:", " refs)
       else
-        oneref
+        List.nth_exn refs
     in
     let mhc = get_mhc_alleles alleles in
     let output_dir =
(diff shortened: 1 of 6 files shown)


src/pipeline_edsl/to_workflow.ml
```diff
@@ -856,20 +856,35 @@ module Make (Config : Compiler_configuration)
     )
 
   let topiary ?(configuration=Tools.Topiary.Configuration.default)
-      vcf predictor alleles =
-    let v = get_vcf vcf in
-    let reference_build = v#product#reference_build in
+      vcfs predictor alleles =
+    let vs = List.map ~f:get_vcf vcfs in
+    let refs = List.map ~f:(fun v -> v#product#reference_build) vs in
```
ihodes: You could use List.dedup here:

```
let refs = ... |> List.dedup in
if List.length refs != 1 then fail ... else we're good
```

src/pipeline_edsl/to_workflow.ml
```diff
@@ -856,20 +856,35 @@ module Make (Config : Compiler_configuration)
...
-      ] in
+    let output_dir =
+      Name_file.in_directory ~readable_suffix:"topiary" Config.work_dir 
+        ([
+          Tools.Topiary.predictor_to_string predictor;
+          Tools.Topiary.Configuration.name configuration;
+          Filename.chop_extension (Filename.basename mhc#product#path);
+        ] @
+          (List.map vs ~f:(fun v ->
+               (Filename.chop_extension (Filename.basename v#product#path))))
```
ihodes: I'd use the full path; otherwise different VCFs from different directories with the same basename could cause issues? I always err on the side of safety here.

src/pipeline_edsl/to_workflow.ml
```diff
@@ -856,20 +856,35 @@ module Make (Config : Compiler_configuration)
...
+        )
+    in
+    let output_file = output_dir // "results.csv" in
```
ihodes: Why not construct an output_file above with the Name_file module, so that you don't need a directory with a single file inside?

armish: Yeah, was thinking about that, but none of the `Name_file` methods pop up to me immediately. For example, how about:

```
let output_file = Name_file.in_directory ~readable_suffix:"topiary.csv" Config.work_dir ...
```

or is there another good way to do this?

src/pipeline_edsl/to_workflow.ml
armish: good catch!

src/pipeline_edsl/to_workflow.ml
armish: Hmm, I definitely like your approach much better. Didn't know about the `dedup` before :\

src/pipeline_edsl/to_workflow.ml
```diff
@@ -858,15 +858,17 @@ module Make (Config : Compiler_configuration)
...
+    in
     let reference_build =
-      if List.exists refs ~f:(fun r -> r <> oneref)
+      if List.length refs > 1
       then
         ksprintf failwith "VCFs do not agree on their reference build: %s"
           (String.concat ~sep:", " refs)
       else
-        oneref
+        List.nth_exn refs
```
ihodes: need a # here :)

src/pipeline_edsl/to_workflow.ml
ihodes: I believe that's the way to do it, would have to double-check/see other examples.

src/pipeline_edsl/to_workflow.ml
ihodes: I didn't either, just hoped that @smondet would've had a List.uniq or equivalent in Nonstd (and, of course, there it was) :)

src/pipeline_edsl/to_workflow.ml
armish: argh!

smondet:
@armish the Travis errors seems to be an actual build error, can you have a look?

armish:
@smondet: done!

PR #436, opened 2017-03-03
Add Picard.reorder_sam

ihodes:
for https://github.com/hammerlab/epidisco/issues/167#issuecomment-283806812

### DIFF
--- a/src/bfx_tools/picard.ml
+++ b/src/bfx_tools/picard.ml
@@ -159,16 +159,14 @@ let reorder_sam
   let tmp_dir =
     Filename.chop_extension output_bam_path ^ ".tmpdir" in
   let input_bam_path = input_bam#product#path in
-  let name =
-    sprintf "picard-reorder_sam-%s" Filename.(basename input_bam#product#path) in
-  let product = transform_bam input_bam#product output_bam_path in
-  let reference =
-    Machine.get_reference_genome run_with
-      begin match reference_build with
-      | None -> input_bam#product#reference_build
-      | Some r -> r
-      end
+  let reference_build = match reference_build with
+  | None -> input_bam#product#reference_build
+  | Some r -> r
   in
+  let reference = Machine.get_reference_genome run_with reference_build in
+  let name =
+    sprintf "picard-reorder_sam-%s"
+      Filename.(basename input_bam#product#path) in
   let fasta = Reference_genome.fasta reference in
   let make =
     let reference_path = fasta#product#path in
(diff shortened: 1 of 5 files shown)


src/bfx_tools/picard.ml
```diff
@@ -150,6 +150,55 @@ let mark_duplicates
...
+    ?reference_build ~input_bam output_bam_path
+  =
+  let open KEDSL in
+  let picard_jar = Machine.get_tool run_with Machine.Tool.Default.picard in
+  let tmp_dir =
+    Filename.chop_extension output_bam_path ^ ".tmpdir" in
+  let input_bam_path = input_bam#product#path in
+  let name =
+    sprintf "picard-reorder_sam-%s" Filename.(basename input_bam#product#path) in
+  let product = transform_bam input_bam#product output_bam_path in
```
smondet: In the case where `reference_build` is `Some nrb`, the new bam should carry the new `nrb` as its ref-build I think.

src/bfx_tools/picard.ml
```diff
@@ -150,6 +150,55 @@ let mark_duplicates
...
+          tmp_dir
+          (Filename.quote input_bam_path)
+          (Filename.quote output_bam_path)
+          reference_path)
+    in
+    Machine.run_big_program ~name run_with program
+      ~self_ids:["picard"; "mark-duplicates"] in
+  workflow_node product
+    ~name ~make ~edges:[
+    depends_on input_bam;
```
smondet: Do we know whether the tools works on unsorted bams? and indexing?

src/pipeline_edsl/semantics.ml
```diff
@@ -173,6 +173,12 @@ module type Bioinformatics_base = sig
     [ `Bam ] repr ->
     [ `Bam ] repr
 
+  val picard_reorder_sam:
+    ?mem_param : string ->
```
smondet: The other tools have a `?configuration` argument which seems cleaner (?)

src/pipeline_edsl/to_workflow.ml
```diff
@@ -731,6 +731,23 @@ module Make (Config : Compiler_configuration)
         ~run_with ~input_bam output_bam
     )
 
+  let picard_reorder_sam ?mem_param ?reference_build bam =
+    let input_bam = get_bam bam in
+    let reference_build_param =
+      match reference_build with
+      | None -> ""
```
smondet: shouldn't it be `input_bam#ref_build`?

src/bfx_tools/picard.ml
ihodes: Per @julia326 it should.

src/pipeline_edsl/semantics.ml
ihodes: I was thinking about it, but for something which has only two params, I thought that would be heavyweight (and easy to introduce later if needed).

ihodes:
@smondet back to you!

PR #437, opened 2017-03-05
Add cDNA for the mm10 genome

armish:
This is needed, for example, when running kallisto against an mm10 sample/reference. Otherwise, any tool asking for the mm10 cDNA breaks the workflow.


ihodes:
LGTM!

PR #438, opened 2017-03-08
Fix build with Cmdliner 1.0.0

smondet:
The change is still compatible with previous versions.

### DIFF
--- a/src/app/input_gen.ml
+++ b/src/app/input_gen.ml
@@ -34,9 +34,10 @@ let cmd () =
     Arg.(value & flag & info ["single-ended"] ~doc)
   in
   Term.(pure (fun sample_name host directory single_ended ->
-      let host = match Ketrew_pure.Host.of_string host with
-      | `Ok host -> host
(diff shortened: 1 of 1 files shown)


src/app/input_gen.ml
```diff
@@ -34,9 +34,10 @@ let cmd () =
     Arg.(value & flag & info ["single-ended"] ~doc)
   in
   Term.(pure (fun sample_name host directory single_ended ->
-      let host = match Ketrew_pure.Host.of_string host with
-      | `Ok host -> host
-      | `Error msg -> printf "Error parsing host.\n%!"; exit 1
+      let host =
+        match Ketrew_pure.Host.of_string host with
+        | `Ok host -> host
+        | `Error msg -> printf "Error parsing host.\n%!"; Pervasives.exit 1
```
ihodes: Curious what was required by Cmdliner here?

smondet: There is a `Term.exit` function in 1.0.0.

PR #446, opened 2017-04-04
Separate bwa configs for mem and aln sub-methods

### LINKED ISSUE #445: Default `bwa mem` penalty parameters in biokepi are
These are the defaults we have right now:

https://github.com/hammerlab/biokepi/blob/master/src/bfx_tools/bwa.ml#L8-L9
```ocaml
(* ... *)
  let default_gap_open_penalty = 11
  let default_gap_extension_penalty = 4
(* ... *)
```

which is compiled into `bwa mem -O 11 -E 4` by default. However, `bwa mem`'s defaults are `-O 6 -E 1` according to [the tool documentation](http://bio-bwa.sourceforge.net/bwa.shtml) (kudos: @smondet)

This discrepancy might discourage `bwa mem` from adding gaps to the alignments since the biokepi ones are relatively higher than the default ones (and these scores are often co-optimized). I am currently experimenting with a few parameter combinations on real patient data to see whether this is actually causing us any trouble and if so, we probably need to convert these back to the `mem` defaults.

armish:
Fixes #445.

I might have done something stupid with the way I use modules/functors since I barely use them. Let me know if the coding style looks off.

### DIFF
--- a/src/bfx_tools/bwa.ml
+++ b/src/bfx_tools/bwa.ml
@@ -4,31 +4,47 @@ open Common
 module Remove = Workflow_utilities.Remove
 
 module Configuration = struct
+  type bwa_cfg_type = {
+    name: string;
+    gap_open_penalty: int;
+    gap_extension_penalty: int;
+    mismatch_penalty: int;
+  }
 
-  let default_gap_open_penalty = 11
-  let default_gap_extension_penalty = 4
-
-  module Common = struct
-    type t = {
-      name: string;
-      gap_open_penalty: int;
-      gap_extension_penalty: int;
-    }
-    let default = {
-      name = "default";
-      gap_open_penalty = default_gap_open_penalty;
-      gap_extension_penalty = default_gap_extension_penalty;
-    }
+  module Bwa_Config (D: sig val default: bwa_cfg_type end) = struct
+    type t = bwa_cfg_type
     let name t = t.name
-    let to_json {name; gap_open_penalty; gap_extension_penalty} =
+    let default = D.default
+    let to_json
+        {name; gap_open_penalty; gap_extension_penalty; mismatch_penalty}
+      =
       `Assoc [
         "name", `String name;
         "gap_open_penalty", `Int gap_open_penalty;
         "gap_extension_penalty", `Int gap_extension_penalty;
+        "mismatch_penalty", `Int mismatch_penalty;
       ]
   end
-  module Aln = Common
-  module Mem = Common
+
+  (* 
+    Defaults transcribed from BWA manual: 
+      http://bio-bwa.sourceforge.net/bwa.shtml
+  *)
+  let bwa_mem_default = {
+    name = "default-bwa-mem";
+    gap_open_penalty = 6;
+    gap_extension_penalty = 1;
+    mismatch_penalty = 4;
+  }
+  module Mem = Bwa_Config(struct let default = bwa_mem_default end)
+
+  let bwa_aln_default = {
+    name = "default-bwa-aln";
+    gap_open_penalty = 11;
+    gap_extension_penalty = 4;
+    mismatch_penalty = 3;
+  }
+  module Aln = Bwa_Config(struct let default = bwa_aln_default end)
 
 end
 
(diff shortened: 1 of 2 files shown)


src/bfx_tools/bwa.ml
```diff
@@ -4,31 +4,47 @@ open Common
...
+        {name; gap_open_penalty; gap_extension_penalty; mismatch_penalty}
+      =
       `Assoc [
         "name", `String name;
         "gap_open_penalty", `Int gap_open_penalty;
         "gap_extension_penalty", `Int gap_extension_penalty;
+        "mismatch_penalty", `Int mismatch_penalty;
       ]
   end
-  module Aln = Common
```
ihodes: I'd keep this unless the config possibilities for both aligners are exactly the same.

src/bfx_tools/bwa.ml
```diff
@@ -4,31 +4,47 @@ open Common
 module Remove = Workflow_utilities.Remove
 
 module Configuration = struct
+  type bwa_cfg_type = {
```
ihodes: could be named `bwa_config` instead (since we know it's a type!)

src/bfx_tools/bwa.ml
```diff
@@ -175,8 +192,9 @@ let align_to_sam
       String.concat ~sep:" " [
         "bwa aln";
         "-t"; Int.to_string processors;
-        "-O"; Int.to_string configuration.Configuration.Aln.gap_open_penalty;
-        "-E"; Int.to_string configuration.Configuration.Aln.gap_extension_penalty;
+        "-O"; Int.to_string configuration.Configuration.gap_open_penalty;
+        "-E"; Int.to_string configuration.Configuration.gap_extension_penalty;
+        "-M"; Int.to_string configuration.Configuration.mismatch_penalty;
```
smondet: :heart: that we use one parameter name for the same option that is carried by 2 different flags for no reason !

src/bfx_tools/bwa.ml
smondet: I don't understand your comment, the module `Aln` is still down there :arrow_down:

src/bfx_tools/bwa.ml
smondet: Yes and full words are always more readable, “config ≥ cfg”.

src/bfx_tools/bwa.ml
ihodes: Good point, thrown off by the loss of the submodule qualification below (added comment)

src/bfx_tools/bwa.ml
```diff
@@ -319,8 +337,9 @@ let mem_align_to_bam
          ~sample_name:(Input_reads.sample_name input)
          ~read_group_id:(Input_reads.read_group_id input));
       "-t"; Int.to_string processors;
-      "-O"; Int.to_string configuration.Configuration.Mem.gap_open_penalty;
```
ihodes: Might be worth keeping this submodule in here (would make it easier to make mem vs. aln specific configs if needed). Probably doesn't matter, though…

src/bfx_tools/bwa.ml
```diff
@@ -4,31 +4,47 @@ open Common
...
-      gap_open_penalty: int;
-      gap_extension_penalty: int;
-    }
-    let default = {
-      name = "default";
-      gap_open_penalty = default_gap_open_penalty;
-      gap_extension_penalty = default_gap_extension_penalty;
-    }
+  module Bwa_Config (D: sig val default: bwa_cfg_type end) = struct
+    type t = bwa_cfg_type
```
smondet: To fix @ihodes issues with module names/qualified-names, I think you can replace this line with:

```ocaml
type t = bwa_cfg_type = {
  name: string;
  gap_open_penalty: int;
  gap_extension_penalty: int;
  mismatch_penalty: int;
}
```

smondet: Or if you don't want the repetition:

```ocaml
module Common_config = struct
  type t = {
    name: string;
    gap_open_penalty: int;
    gap_extension_penalty: int;
    mismatch_penalty: int;
  }
end

module Bwa_Config (D: sig val default: Common_config.t end) = struct

   include Common_config
   
...
```

ihodes: Is there a reason to not just define `t` inside of `Bwa_config` directly?

smondet: @ihodes we need it for the parameter of the functor

armish:
@ihodes @smondet: thanks a lot for the suggestions! I just incorporated those so should be ready to go whenever you are.

src/pipeline_edsl/common_pipelines.ml
```diff
@@ -48,10 +48,11 @@ module Somatic = struct
       pair ~normal ~tumor in
     let bam_pairs =
       let non_default =
-        { Bwa.Configuration.Aln.
-          name = "config42";
+        let open Bwa.Configuration.Common_config in
```
smondet: I would open the right one here (`.Aln`) not the `Common_config`.

armish: ah, good catch! This is fixed now.

PR #448, opened 2017-04-11
STAR aligner shouldn't do sorting; does in-memory

ihodes:
Our tools with automatically sort the BAM if needed

### DIFF
--- a/src/bfx_tools/star.ml
+++ b/src/bfx_tools/star.ml
@@ -130,13 +130,13 @@ let align
   let star_index = index ~reference_build ~run_with in
   let reference_dir = (Filename.dirname reference_fasta#product#path) in
   let star_index_dir = sprintf "%s/star-index/" reference_dir in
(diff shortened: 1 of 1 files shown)


src/bfx_tools/star.ml
```diff
@@ -131,12 +131,12 @@ let align
   let reference_dir = (Filename.dirname reference_fasta#product#path) in
   let star_index_dir = sprintf "%s/star-index/" reference_dir in
   (* STAR appends Aligned.sortedByCoord.out.bam to the filename *)
-  let result = sprintf "%sAligned.sortedByCoord.out.bam" result_prefix in
+  let result = sprintf "%sAligned.out.bam" result_prefix in
```
smondet: the comment above the line is out of date now, right?

smondet: I guess the comment should be now saying that this is the observed filename with `--outSAMtype BAM Unsorted`

ihodes: Good catch! Fixed.

PR #450, opened 2017-04-19
Update build

smondet:
Makes Biokepi build with https://github.com/hammerlab/ketrew/pull/529
(backwards compatible)


smondet:
I changed https://github.com/hammerlab/ketrew/pull/529 so this is not useful anymore.

PR #452, opened 2017-04-21
First pass of "save" functionality

ihodes:


tavinathanson:
What would be the best way to support `Bfx.save` writing to `gs://` URLs?

@smondet I see the [gsutil_cp function](https://github.com/hammerlab/biokepi/blob/3e1cb08f05bea150ab794a0f65265e160289b6d4/src/run_environment/workflow_utilities.ml#L289); is there an existing function that knows to write either locally or via `gsutil`?

tavinathanson:
From @smondet:

> @tavinathanson it wouldn't be hard to implement (just look at the implementation of `Bfx.input`), then I'm not sure; maybe better doing `Bfx.save` to a directory and then, separately (throught Ketrew and the end of the pipeline or not), do `gsutil rsync` of the whole thing

tavinathanson:
Probably would be a separate PR then?

ihodes:
@smondet @tavinathanson I think it's best we merge this in, then I can file some issues for 

1. supporting writing to gs://
2. generating a csv manifest
3. generating the .json files alongside the saved products

Does that seem reasonable?

tavinathanson:
@ihodes works for me if @smondet is cool with the code

tavinathanson:
@ihodes would love to use this in master, thoughts on getting it in?

smondet:
@ihodes Travis says that you didn't update the `src/examples/`:

````
module My_cluster = struct

  let max_processors = 42

  let work_dir = "/work/dir/"

  let results_dir = "/results/dir/"

  let datasets_home = "/datasets/"

  let machine =

    Biokepi.Setup.Build_machine.create "ssh://example.com/tmp/KT/"

end

File "./src/examples/edsl_extension_register_result.ml", line 269, characters 7-182:

Error: Signature mismatch:

       ...

       The value `results_dir' is required but not provided
````

File: src/bfx_tools/optitype.ml [hunk 1]
```diff
@@ -4,8 +4,25 @@ open Common
 
 
 type product = <
-    is_done : Ketrew_pure.Target.Condition.t option ;
-    path: string >
+  host: Ketrew_pure.Host.t;
+  is_done : Ketrew_pure.Target.Condition.t option ;
+  path: string >
+
+let transform_optitype_product ?host ~path o =
```
smondet: since the word “transform” is very vague it would be worth documenting those functions.
(and maybe `move_*_product` or `displace_*_product`?)

src/pipeline_edsl/semantics.ml
```diff
@@ -47,6 +47,11 @@ module type Bioinformatics_base = sig
      URL. E.g. ["https://data.example.com/my-sample.bam?filename=sample-from-example.bam"].
   *)
 
+  val save :
```
smondet:
```
(** Tag a product of the EDSL as “interesting to be saved” (the meaning is implementation dependent; when compiling to a Ketrew workflow this means copying the file(s) to another directory along with a JSON reprensentation of their provenance. *)
```

src/pipeline_edsl/to_json.ml
```diff
@@ -57,6 +56,10 @@ module Make_serializer (How : sig
       "URL", u;
     ]
 
+  let save ~name thing =
+    fun ~var_count ->
+      function_call name (["name", string name])
```
smondet: the first `name` should be `"save"` there?

src/pipeline_edsl/to_workflow.ml
```diff
@@ -231,14 +231,21 @@ end
...
+
   val map_reduce_gatk_indel_realigner : bool
+  (** Whether or not to scatter-gather the indel-realigner results. *)
 
+  val input_files: [ `Copy | `Link | `Do_nothing ]
   (** What to do with input files: copy or link them to the work directory, or
       do nothing. Doing nothing means letting some tools like ["samtools sort"]
       write in the input-file's directory. *)
-  val input_files: [ `Copy | `Link | `Do_nothing ]
 end
```
smondet: I think `results_dir` should be a `string option` so that we can put `None` (and do nothing) in the defaults below:
:arrow_down:

src/pipeline_edsl/to_json.ml
ihodes: 👍

src/pipeline_edsl/semantics.ml
ihodes: 👍

ihodes:
This is ready to merge, I think! /cc @smondet

tavinathanson:
@ihodes @smondet would love to use this in master today, so I vote for addressing anything else in a follow-up.

PR #465, opened 2017-05-10
Use local tmp dir for NetMHC family

armish:
Since the one on the mounted folder seems to be causing issues from time to time (reported by @julia326).

### DIFF
--- a/src/environment_setup/netmhc.ml
+++ b/src/environment_setup/netmhc.ml
@@ -110,8 +110,15 @@ let guess_folder_name tool_file_loc =
 (* 
   netMHC tools will be populating this folder,
   so this is an important "tmp" folder!
+
+  And it turns out that having this on a mounted
+  folder is not that effective since NetMHC
+  occasionaly fails to create/remove new folders
+  here if the resources are limited (e.g. NFS).
+  So make sure you know where your parent_dir is.
 *)
-let tmp_dir install_path = install_path // "tmp"
+let tmp_dir parent_dir = parent_dir // "tmp"
+let local_tmp_dir = tmp_dir "/"
 
 (* 
   NetMHC tools play nicely against Python 2.x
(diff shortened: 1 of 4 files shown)


smondet:
I don't think `/tmp` should be set in stone there without being configurable in some way, very often biokepi nodes run on machines that have very little local storage (and bfx tools that use temp-files usually, it's because tmp-data doesn't fit in memory, e.g. `samtools sort`).

julia326:
With NetMHC tools though, it seems like they don't play nice with anything but local storage - what do you think we should do in this case?

smondet:
@julia326 @armish I just think it should be somewhat configurable, from the biokepi-machine for example (even if `/tmp` is the default, it's just that we've  been bitten by these things quite a few times).

Also, is there an issue reported somewhere about NetMHC not behaving posixly?

ihodes:
I believe this works (solves the tmp dir issue + seems to have allowed NetMHC to run correctly!)

armish:
Phew - glad this solved your issues!

@julia326: I don't think we have an issue describing this problem but I am not sure where would be the best place to put this information. Maybe mentioning this in the bundle README will help others to not hit this in the future (and most importantly without noticing)?

In terms of allowing this to be customized: I do agree that this should be customizable and definitely on the biokepi level. I am not sure whether we should surface yet another `ENV` variable to the epidisco and alike. @smondet @ihodes: what do you think?

julia326:
@armish I can put an issue into biokepi, that seems like the best place for it. Can also add a note to the netmhc-bundle README, though luckily in this case the netMHC error was informative enough that it became pretty clear what the problem is.

src/environment_setup/netmhc.ml
```diff
@@ -24,6 +24,38 @@ type netmhc_file_locations = {
...
+  occasionaly fails to create/remove new folders
+  here if the resources are limited (e.g. NFS).
+  So make sure you know where your parent_dir is.
+*)
+let local_tmp_dir = "/tmp/" (* default: /tmp *)
+
+type netmhc_config = {
+  netmhc_tmpdir: string;
+  netmhc_file_locations: netmhc_file_locations;
+}
```
smondet: :heart:

armish:
@smondet: just came here to say this is ready for your review and 😍

PR #468, opened 2017-05-17
Condaify all the tools

armish:
The idea here is to install as many tools as possible via conda's bioconda channel and worry less about all the weird setup/dependency issues.

Currently testing OptiType first (cc: @tavinathanson and @rleonid)


tavinathanson:
@armish super awesome! I'll try it out!

tavinathanson:
Successfully ran this on 75 samples (after a few retries)! And made some additions in https://github.com/hammerlab/biokepi/pull/468/commits/f89933344c6e9908ddfbae7795d2c8acdc50e973.

Will defer to @rleonid on the remainder of the review.

src/environment_setup/conda.mli
```diff
@@ -1,13 +1,20 @@
 open Biokepi_run_environment
 
+type python_version_type = [
+  | `Python2
+  | `Python3
+  | `PythonCustom of string
+  | `PythonToolDependency of string
+]
+
```
rleonid: Why not use a regular variant? I would also document this type, as the meaning these variants isn't obvious.

smondet: `+1` for the documentation

I don't mind the polymorphic variant Vs regular variant.

The guideline is to use underscores and never CamelCase (so → `` `Python_tool_dependency``)

armish:
Never mind my latest push. Was just for @tavinathanson to test something from this branch. This is almost done but need a few `rebase`s and `squash`s, and also some documentation on all the major additions/changes.

Stay tuned!

src/run_environment/hla_utilities.ml
```diff
@@ -0,0 +1,97 @@
...
+      (hlarp_result#render#name)
+  in
+  let edges = [ depends_on hlarp_result; ] in
+  let product = single_file ~host:(Machine.as_host run_with) output_path in
+  let tmp_path = output_path ^ ".tmp" in
+  let make = Machine.(
+    run_program run_with
+      ~requirements:[ `Quick_run; ]
+      Program.(
+        shf "cat %s \
```
tavinathanson: I thought the point of `hlarp` was to avoid needing to do munging like this?

@rleonid can `hlarp` offer this is an output format?

src/environment_setup/conda.mli
armish: regular variant it is then. I will give revisit this whole branch once all the implementation is done and will make sure I document all the new parts thoroughly.

Switching between language contexts has this weird effect of involuntarily mixing/matching coding conventions on me. Good catch - will also address current styling issues once all is done.

src/run_environment/hla_utilities.ml
armish: @tavinathanson: yep - this is the controversial code that made fun of during the retreat and also the same exact thing @rleonid mentioned here: https://github.com/hammerlab/discohorts/issues/9

I just factored it out from the hlarp module since this was mostly about making it compatible with `mhctools`-family of libraries. I have another experimental `hlarp` branch that I am currently working on which will allow printing plain lists instead of annotated CSVs and it will also enable have a subcommand to extract the consensus calls across multiple typing runs when we have more than one tool/data-type for a single patient.

Once that is done, I will get rid of this bash foo and will save @ihodes from all the headache ;) Hopefully cohorts/discohorts will also benefit from those additional hlarp functions.

armish: that piece of code does stand out, doesn't it? :)

rleonid: `hlarp`'s chosen output format was something agreed upon that the Python sweet of tools should support.

armish:
(belated) thank you so much for all the help with getting this branch up to speed and merged, @smondet! Hope these changes won't cause too much of a trouble for the epidico folks.

PR #472, opened 2017-05-26
Call vaxrank from the `output_folder` instead of the default `root`

armish:
to work around this issue: https://github.com/hammerlab/vaxrank/issues/130


ihodes:
LGTM, though I've had many Vaxrank successes, so I'm wondering what else might be going on?

PR #474, opened 2017-06-10
Make MuTect v1 great again

armish:
@julia326 found out that mutect v1 has started failing since @smondet's recent migration of biokepi runners from Java 1.7 to 1.8 (c.f. https://github.com/hammerlab/secotrec/issues/64). This is apparently a well-known issue and has been acknowledged by Broad. 

This change wraps the mutect tool with a conda environment that comes with `openjdk-7.x` to make sure it keeps running no matter what the global Java version is.

(Tested and confirmed that this works as intended)


hammer:
👍🏼 for the fix (makes me think we should bundle a fixed runtime with all log-running tools), 👎🏼 for Trump reference (requires trigger warning)

hammer:
Gah meant long-running of course

PR #478, opened 2017-07-05
Provenance on steroids

smondet:
Still to do:

- [x] use the `Provenance_description.t` to output a `.json` (and `.csv`?), to `save`d (or all?) files.
- [x] propagate the change to make sure extensions of the EDSL are somewhat easy to update.

This has already the advantage of keep around the JSON serialization of the configurations (and not just their name).

### DIFF
--- a/src/pipeline_edsl/to_workflow.ml
+++ b/src/pipeline_edsl/to_workflow.ml
@@ -920,7 +920,7 @@ module Make (Config : Compiler_configuration)
       begin match get_list (AF.get_file l) with
       | Fastq one :: [] ->
         Fastq one
-        |> AF.with_provenance "concat" ["element_1", AF.get_provenance l]
+        |> AF.with_provenance "concat" ["single-element", AF.get_provenance l]
       | Fastq first_fastq :: _ as lfq ->
         let fqs = List.map lfq ~f:get_fastq in
         let r1s = List.map fqs ~f:(KEDSL.read_1_file_node) in
--- a/src/pipeline_edsl/to_workflow.ml
+++ b/src/pipeline_edsl/to_workflow.ml
@@ -261,6 +261,16 @@ module Provenance_description = struct
     string_arguments: (string * string) list;
     json_arguments: (string * Yojson.Basic.json) list;
   }
+  let rec to_yojson t : Yojson.Basic.json =
+    let fields =
+      List.concat [
+        List.map t.sub_tree_arguments ~f:(fun (k, v) -> k, to_yojson v);
+        List.map t.string_arguments ~f:(fun (k, v) -> k, `String v);
+        t.json_arguments;
+      ]
+    in
+    `Assoc (("node-name", `String t.name) :: fields)
+
 
 end
 module Annotated_file = struct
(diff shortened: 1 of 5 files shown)


src/pipeline_edsl/to_workflow.ml
```diff
@@ -255,41 +253,116 @@ module Defaults = struct
   let results_dir = None
 end
 
+module Provenance_description = struct
+
+  type t = {
+    name: string;
+    sub_tree_arguments: (string * t) list;
```
armish: I see how we will make use of the `string` and `json` arguments, but what is the `sub_tree` for?

smondet: `string` and `json` are the "contants" (like reference-build, configuration, etc.) and the sub-tree is the provenance of the arguments of the EDSL function, e.g the input VCF here:

```ocaml
  let vcf_annotate_polyphen vcf =
    let v = get_vcf (AF.get_file vcf) in
    let output_vcf = (Filename.chop_extension v#product#path) ^ "_polyphen.vcf" in
    let reference_build = v#product#reference_build in
    Vcf (
      Tools.Vcfannotatepolyphen.run ~run_with ~reference_build ~vcf:v ~output_vcf
    )
    |> AF.with_provenance "vcf-annotate-polyphen" ["vcf", AF.get_provenance vcf]
```

smondet:
@armish this is ready for review, cf. also https://github.com/hammerlab/epidisco/pull/196

src/test/ttfi_pipeline.ml
```diff
@@ -120,7 +120,7 @@ module Run_test(Test_pipeline : TEST_PIPELINE) = struct
       let module Ketrew_pipeline = Test_pipeline(Workflow_compiler) in
       let workflow =
         Ketrew_pipeline.run ()
-        |> Biokepi.EDSL.Compile.To_workflow.File_type_specification.get_unit_workflow
+        |> Biokepi.EDSL.Compile.To_workflow.get_workflow
```
armish: Ah - now I understand why the new provenance addition is important. Cool!

src/pipeline_edsl/to_workflow.ml
```diff
@@ -529,14 +667,26 @@ module Make (Config : Compiler_configuration)
         ~name:(sprintf "Input-bam: %s" sample_name)
         ~edges:[depends_on f]
     )
+    |> AF.with_provenance "bam" ["file", AF.get_provenance input]
```
armish: So, are we going to use `file` for everything before they fed into to DSL? The way you use these keywords make sense but I can imagine them giving a hard time for somebody who would like to extend this with little knowledge of the framework?

Nothing urgent to address here but maybe we can think about ways to make these more accessible to a newcomer as part of another issue?

smondet: We have been using `Bfx.file` for a while. 
It is usually hidden by the higher-level functions of the edsl-library:
https://github.com/hammerlab/biokepi/blob/provenance-on-steroids/src/pipeline_edsl/pipeline_library.mli#L91

